PR #126, opened 2025-12-17
Kan 150/file path to a non existant file crashes the app

fulsomenko:
Fixes persistence initialization and file watching issues:

- Fix initial file template to use proper V2 format with metadata
- Remove redundant version fields from PersistenceMetadata
- Add instance ID check to file watcher to prevent false positives
- Add migration verification with automatic backup cleanup

## What

Three related persistence improvements:
1. Creating new board files now generates proper V2 format with instance_id and saved_at metadata
2. Simplified file format by removing redundant `format_version` and `schema_version` fields
3. File watcher now ignores own writes via instance ID comparison
4. V1→V2 migration now verifies success and auto-removes backup

## Why

**Problem 1**: Running `kanban test.json` on a non-existent file crashed with "missing field `instance_id`" error because the initial template used legacy V1 format without metadata.

**Problem 2**: The file format had three version indicators which were redundant:
- `version` (top level) - actually validated during load
- `metadata.format_version` - duplicate of `version`
- `metadata.schema_version` - never checked anywhere

**Problem 3**: Creating boards triggered "external file change detected" dialog because the file watcher's pause/resume mechanism had a race condition, causing the app to detect its own saves as external changes.

**Problem 4**: V1→V2 migration left `.v1.backup` files around even after successful migration, cluttering the filesystem without providing value.

## How

**Initial file template fix:**
- Updated template to V2 format with complete structure
- Added uuid and chrono dependencies to kanban-cli for metadata generation
- Template now includes version, metadata (instance_id, saved_at), and data sections

**Version field cleanup:**
- Removed `format_version` and `schema_version` from PersistenceMetadata struct
- Updated initial file template and migration to not add redundant fields
- Updated all tests and documentation to reflect simplified structure
- Backward compatible: Serde ignores unknown fields, so old files still load

**File watcher fix:**
- Added instance ID check in file change handler (app.rs:1699-1718)
- When file change detected: loads metadata and compares instance_id
- If IDs match: ignore change (it's our own write)
- If load fails: falls through to external change logic (fail-safe)
- Independent of timing, provides bulletproof filtering

**Migration verification:**
- Added `verify_migration()` method that validates:
  - Migrated file is valid JSON
  - Has correct `version: 2` field
  - Has valid `metadata` object
  - Data exactly matches original V1 data
- After migration: verification runs automatically
- Success: backup removed, logs "Migration verified, backup removed"
- Failure: backup preserved, logs error with backup location

## Testing

All 214 tests pass across all crates:
- kanban-domain: 41 tests
- kanban-persistence: 20 tests (including updated migration test)
- kanban-tui: 83 tests
- Integration tests: All passing

**Manual testing:**
- Created new board files - proper V2 format generated, no crash
- Tested board creation - no false "external change detected" dialogs
- Tested V1→V2 migration - backup removed after successful verification
- Confirmed external changes from other instances still trigger dialog correctly
- Verified old files with redundant fields still load correctly (backward compatible)

### DIFF
--- a/.changeset/kan-150-file-path-to-a-non-existant-file-crashes-the-app.md
+++ b/.changeset/kan-150-file-path-to-a-non-existant-file-crashes-the-app.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+---
+
+- feat: Add migration verification and automatic backup cleanup
+- fix: Add instance ID check to file watcher to prevent false positives
+- fix: Remove redundant version fields from PersistenceMetadata
+
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -735,6 +735,7 @@ dependencies = [
  "clap",
  "kanban-core",
  "kanban-domain",
+ "kanban-persistence",
  "kanban-tui",
  "tokio",
  "tracing",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -731,6 +731,7 @@ name = "kanban-cli"
 version = "0.1.13"
 dependencies = [
  "anyhow",
+ "chrono",
  "clap",
  "kanban-core",
  "kanban-domain",
@@ -738,6 +739,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "uuid",
 ]
 
 [[package]]
--- a/crates/kanban-cli/Cargo.toml
+++ b/crates/kanban-cli/Cargo.toml
@@ -23,3 +23,5 @@ tokio.workspace = true
 anyhow.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
+uuid.workspace = true
+chrono.workspace = true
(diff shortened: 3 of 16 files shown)


fulsomenko:
## Performance Consideration: File Watcher Instance ID Check

I've reviewed the file watcher implementation (app.rs:1699-1718) and wanted to discuss the double I/O pattern:

**Current Flow:**
1. File watcher detects change (OS notification)
2. Load file to read instance_id from metadata
3. Compare instance_id to filter own writes

**Analysis:**

**Why this approach works well:**
- **Correctness over performance**: The previous pause/resume mechanism had race conditions. Reading instance_id provides reliable detection.
- **OS page cache**: File reads are typically served from OS cache after the first read, so the second load is often a memory operation, not disk I/O.
- **Rare event**: External file changes are infrequent in normal usage - this only triggers on actual external modifications.

**Potential optimizations (if profiling shows this is a bottleneck):**
1. **Metadata-only reads**: Platform-specific APIs could read just the first ~200 bytes (metadata section)
2. **Cached instance ID**: Store the last-read instance_id to avoid file access on every event
3. **Streaming JSON parser**: Parse only the metadata field without deserializing entire file

**Recommendation:** Keep the current implementation for correctness and simplicity. Only optimize if real-world profiling shows performance impact with large board files.

The trade-off heavily favors correctness here - race-free operation is more valuable than micro-optimizing an infrequent event.

PR #128, opened 2025-12-17
Kan 151/kanban cli

fulsomenko:
Adds comprehensive CLI interface for kanban operations:

- Board, column, card, and sprint CRUD operations
- Bulk operations (archive, move, assign-sprint)
- Export/import functionality
- Shell completion support (bash, zsh, fish, powershell)
- Git integration commands (branch-name, git-checkout)

**What**: `kanban-cli` crate providing a subcommand-based CLI interface that mirrors TUI functionality, enabling scripting and automation of kanban operations.

**Why**: Enable programmatic access to kanban boards for automation, CI/CD integration, and users who prefer CLI over TUI.

**How**: 
- Implements `KanbanOperations` trait for CLI context
- Uses clap for argument parsing with subcommands
- JSON output format for machine parsing
- Positional file argument or `KANBAN_FILE` env var for data file

**Testing**: 36 integration tests covering all commands (`cargo test --package kanban-cli`)


fulsomenko:
## Follow-up Notes

### Sprint Status Validation
The sprint status validation (e.g., preventing assignment to non-active sprints) is a pre-existing issue that affects both the TUI and CLI. This is not introduced by this PR and should be addressed separately if desired.

### Dry-run Mode
A `--dry-run` flag for bulk operations (`bulk-archive`, `bulk-move`, `bulk-assign-sprint`) would be a nice enhancement for a future release. This would allow users to preview what changes would be made before committing them.

### `kanban init --name` Removal
The `kanban init --name` command was removed as part of this CLI refactor. This is an acceptable regression since:
- The command wasn't being actively used
- The same functionality is available via `kanban board create --name "Board Name"`
- The new approach is more consistent with the subcommand pattern used throughout the CLI

crates/kanban-cli/src/context.rs
```diff
@@ -0,0 +1,575 @@
...
+        let cmd = CreateBoard { name, card_prefix };
+        self.execute(Box::new(cmd))?;
+        Ok(self.boards.last().unwrap().clone())
+    }
+
+    fn list_boards(&self) -> KanbanResult<Vec<Board>> {
+        Ok(self.boards.clone())
+    }
+
+    fn get_board(&self, id: Uuid) -> KanbanResult<Option<Board>> {
```
fulsomenko, line 181: **Performance consideration**: The list operations clone all data on each call (`self.boards.clone()`, `self.columns.clone()`, etc.). For large datasets, this could be a concern. Consider returning references or iterators if callers don't need ownership, or implementing pagination for large collections.

This is acceptable for the CLI use case where data is typically accessed once per command invocation, but worth noting for future optimization if needed.

crates/kanban-cli/src/handlers/card.rs
```diff
@@ -0,0 +1,228 @@
...
+            let branch = ctx.get_card_branch_name(id)?;
+            output::output_success(serde_json::json!({"branch_name": branch}));
+        }
+        CardAction::GitCheckout { id } => {
+            let cmd = ctx.get_card_git_checkout(id)?;
+            output::output_success(serde_json::json!({"command": cmd}));
+        }
+        CardAction::BulkArchive { ids } => {
+            let count = ctx.bulk_archive_cards(ids)?;
+            ctx.save().await?;
```
fulsomenko, line 91: **Performance consideration**: The bulk operations call `ctx.save().await` once after all operations complete. This is efficient as it avoids multiple disk writes. However, if a failure occurs partway through, some operations will succeed and be saved while others won't.

For transactional behavior (all-or-nothing), the operations would need to be validated first before applying any changes. The current behavior is reasonable for CLI usage where partial success with detailed failure reporting is acceptable.

fulsomenko (commented):
Added performance consideration comments as requested. These are informational notes for future reference, not blocking issues.

PR #151, opened 2025-12-23
KAN-159/implement-sqlite-storage-backend

fulsomenko:
Adds SQLite as a pluggable storage backend alongside existing JSON file storage:

- Add `Database` error variant to kanban-core
- Create SQLite schema with tables for boards, columns, cards, sprints, archived_cards
- Implement `SqliteStore` with `PersistenceStore` trait using sqlx
- Add JSON to SQLite migration utilities
- Make `StateManager` generic over storage backends with auto-detection by file extension

**What**: SQLite storage backend implementing the `PersistenceStore` trait, enabling pluggable storage.

**Why**: JSON file storage becomes O(n) for both read and write as data grows. SQLite provides incremental saves (only changed rows written) and better scalability past ~1000 cards.

**How**: 
- `SqliteStore` implements `PersistenceStore` trait with upsert patterns for incremental writes
- Storage backend auto-detected from file extension (`.db`/`.sqlite` → SQLite, `.json` → JSON)
- Feature flag `--features sqlite` to enable SQLite support
- Connection pooling via sqlx for concurrent access

**Testing**: 
- Unit tests for `SqliteStore` save/load operations
- Migration tests for JSON → SQLite conversion
- Full test suite passes with `cargo test --features sqlite`
- Manual testing with both `.json` and `.db` files


fulsomenko:
## Code Review: SQLite Storage Backend

### Overview

This PR adds SQLite as a pluggable storage backend alongside the existing JSON file storage. The implementation is well-architected and follows the project's SOLID principles with clean trait-based polymorphism.

**Stats**: +2,249 / -19 lines across 14 files

---

### Strengths

- **Clean architecture**: `Arc<dyn PersistenceStore + Send + Sync>` enables seamless backend switching
- **Feature flag approach**: SQLite is opt-in via `--features sqlite`, keeping the default binary lean
- **Proper SQL schema**: Good use of indexes, foreign keys with cascade deletes, and WAL mode for performance
- **`Table` enum pattern**: Using `const fn` for SQL strings avoids string concatenation and SQL injection risks
- **Atomic transactions**: All writes happen in a single transaction with proper ordering (parents before children)
- **Comprehensive tests**: Migration and store tests cover key scenarios

---

### Critical Issue: File Watching Not Supported for SQLite

The current `FileWatcher` implementation won't work with SQLite:

1. **`FileWatcher` is explicitly JSON-focused** (`file_watcher.rs:15-18`):
   ```rust
   /// This implementation is currently designed for single-file JSON persistence
   ```

2. **SQLite uses WAL mode** (`schema.sql:6`):
   ```sql
   PRAGMA journal_mode=WAL;
   ```
   With WAL mode, writes go to `file.db-wal` first, not the main `.db` file. The `FileWatcher` only watches the main file path.

3. **Impact**: If two instances are open (or another process modifies the database), the TUI won't detect external changes and reload. Users will see stale data or get conflict errors on save.

**Suggested solutions** (can be follow-up work):
- Watch all three SQLite files (`.db`, `.db-wal`, `.db-shm`)
- Poll the `metadata.saved_at` column periodically
- Use SQLite's `PRAGMA data_version` to detect changes
- Document as a known limitation when using SQLite

---

### Other Suggestions

#### Minor Issues

1. **Schema evolution not addressed**: The schema has `schema_version = 1` but no migration path when the schema changes in future versions. Consider documenting the approach.

2. **Connection pool size hardcoded** (`sqlite_store.rs:92`):
   ```rust
   .max_connections(5)
   ```
   For a single-user TUI app, this seems excessive. Consider making it configurable or reducing to 1-2.

3. **Extension check is case-sensitive** (`state/mod.rs`):
   ```rust
   match extension {
       "db" | "sqlite" => ...
   ```
   Should handle `.DB`, `.SQLite` etc.

4. **Changeset version**: Listed as `patch` but this is a new feature - consider `minor` version bump.

#### Performance Considerations

5. **Full table scans for sync**: The `sync_table` method fetches all existing IDs then computes the diff. For very large datasets this could be slow - consider tracking dirty/deleted items separately in future.

6. **JSON serialization round-trip**: Data still goes through JSON (`StoreSnapshot.data`) even in SQLite mode. This preserves compatibility but negates some SQLite performance benefits.

#### Missing Items

7. **No CLI command for migration**: The `migrate_json_to_sqlite` utility exists but isn't exposed via CLI. Users would need to write custom code to migrate.

---

### Test Coverage

**Included**:
- Unit tests for `SqliteStore` save/load
- Migration tests (JSON → SQLite)
- Auto-migration behavior tests

**Missing** (can be follow-up):
- Conflict detection between SQLite instances
- Large dataset performance tests
- Concurrent access scenarios

---

### Summary

This is a solid implementation that extends the persistence layer cleanly. The main gap is file watching support for SQLite, which should either be addressed or documented as a known limitation before merge.

**Recommendation**: Request the file watching limitation be documented in the README/PR description, then approve.

PR #247, opened 2026-05-09
windows/fix-keyevent-handling

amanda-amy-frost:
## What

Fixes duplicated key presses from being generated on Windows.

- Add Windows-only check to only respond to `crossterm::event::KeyEventKind::Press` events in `kanban-tui/src/events.rs`

## Why

I have a Windows-centric workflow (for the time being), and I happened to run into this issue after installing `kanban-cli 0.4.1`. I tried various troubleshooting steps that I'll leave out for brevity, but I found the root cause and decided it would be worth a time-boxed amount of effort to try and fix, as I'd like to be able to use this in Windows Terminal. (The issue affects all terminals in Windows.) The TUI would've been unusable for Windows users otherwise, as besides the fact that all text input gets duplicated (and backspace deletes 2 characters at a time), this also breaks the UI, such as not being able to keep the help menu open.

## How

Thankfully, this is a solved issue in crossterm (see the links below), and the logic of this project made it so that the fix only needed to be applied in one place. The fix is a Windows-only check using `cfg` to skip key events that aren't of kind `Press`.

https://github.com/veeso/tui-realm/issues/54
https://github.com/ratatui/ratatui/issues/347
https://docs.rs/crossterm/latest/crossterm/event/struct.KeyEvent.html

## Testing

I did not add any automatic tests, in part because I'm new to Rust and not sure how best to add such a specific thing with limited time, and in part because the change is most easily tested manually, which is what I did.

I started by testing and building and all the other necessary `cargo` commands using the latest commit of the `develop` branch. I tested using `cargo`/`rustc` 1.95.0 on both Windows and Linux (Debian 13).

### On Windows

I first ran tests on the `develop` branch to see if any tests were failing. Indeed, 2 tests fail on Windows:

failures:

```rust
---- config::tests::test_dto_from_config_with_data_file_shows_storage_fields stdout ----

thread 'config::tests::test_dto_from_config_with_data_file_shows_storage_fields' (38844) panicked at crates\kanban-service\src\config.rs:687:9:
got: Some("...coding\\kanban\\crates\\kanban-service\\kanban.json")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- config::tests::test_dto_from_explicit_config_preserves_values stdout ----

thread 'config::tests::test_dto_from_explicit_config_preserves_values' (31420) panicked at crates\kanban-service\src\config.rs:712:9:
got: Some("...\\coding\\kanban\\crates\\kanban-service\\kanban.sqlite")


failures:
[...]

PR #255, opened 2026-05-11
editor/improve-robustness

amanda-amy-frost:
## What

- Refactor the editor in `kanban-tui` so that it can handle an arbitrary `EDITOR` string in a shell-aware and OS-agnostic way.
- Add unit tests to validate rewritten logic.
- Document that VS Code is known to have issues, with enough info for users to know they must use a different `EDITOR`.

### Cargo checks

`cargo fmt`, `build`, `test`, `clippy` all succeed. `clippy` still shows the pre-existing linting issues from #247, but no new issues have been introduced. `cargo test` was run on Windows and Debian 13 Linux (not WSL). No changes from those already reported in #247 (which I know you have a TODO for 😊).

### New dependencies (kanban-tui crate)

See more in "How" section below.

- [shell-words](https://crates.io/crates/shell-words): command-line parsing
- [which](https://crates.io/crates/which): cross-platform `which` in Rust

### Documentation

Take the README and changeset as initial suggestions, as you should obviously decide if/how you want these things to be documented :p 

**EDIT:** I changed the README in a new commit to make it overall better I think. You can see the previous commit to see how it used to be phrased. It's still an initial suggestion that you should change how you see fit. I'm also making an implicit assumption that you have a Mac that you test on for the line I added in the README about being cross-platform.

### Changed behavior for editor fallback

Instead of cycling through a pre-defined vector of fallback editors based on OS, I simplified the failure states to only `notepad` and `vi` for Windows and non-Windows respectively. If you'd prefer to keep the list of fallback editors, I can add that back in (with proper testing and all the rest).

The risk with my change is that someone's existing workflow might become broken, where they have an invalid `EDITOR` that has silently failed or failed in a way they are annoyed but fine with (and/or don't know how to fix). For example, say a user uses Neovim, but their `EDITOR` is incorrect, and the editor in `kanban` has defaulted to `vim` as a result, which may have worked well enough for them. With this change, `vi` will instead be called, and if for some reason `vi` doesn't alias to `vim` or is otherwise broken, then that could break the editor experience for the user until they fix their `EDITOR`.

Since the problem/fix for the above is:

1. Easy to apply (fix `EDITOR`)
2. Very unlikely to affect anyone right now
3. Users *should* discover the above rather than relying on undocumented backup behavior

That's why I find it an acceptable change. But I'm not sure if you'd find this acceptable as a `patch`, so let me know if I need to adjust anything and how :p 

Also, it could be worthwhile to document the default `vi`/`notepad` behavior. Not necessarily right now, but someone might run into it in the future and want to know what's going on. I'm not even sure if it was triggering properly before this PR, because when I was experimenting with VS Code with my EDITOR set to `code`, I got the "Editor: 'code' not found..." error. I had to fully unset EDITOR to get notepad to launch on Windows. That is no longer the case with this PR.

### VS Code remains broken, but with some progress

`code --wait` can now open instead of erroring out (tested on Windows). However, VS Code, does not work well with editing a temporary file. As-is, temporary files show as "not found". A few hacky attempts to fix this have only resulted in being able to open the temporary file, but saving the file does not persist the changes. Moreover, only one file can be opened the first time when launching VS Code from a fresh run of `kanban`. Any attempt to edit the same content or another field on the same card (or another card, etc.) reverts to the "file not found" issue.

See below for more details.

VS Code file not found error for temporary file

## Why

I wanted to originally get VS Code to work as my `EDITOR`, and after some simple hacks on the existing code, I decided to refactor all the logic outside of `edit_in_external_editor()` in order to make the overall approach more generic. I wanted to be able to handle any valid arbitrary `EDITOR` on any system without issue.

Vim (or vim-like editors) were already working on Windows before this change, so it doesn't meaningfully add any new functionality. However, other editors outside of the hardcoded ones specified in the previous version are more likely to work now.

## How

As I have no idea how `edit_in_external_editor()` works in terms of specifics, I wanted to leave that code as unchanged as possible. It also makes the diff easier to read. My original problem was just getting VS Code to open, so I wasn't thinking of running into issues with the temp file handling, but that's Microsoft products for you.

In any case, I knew I wanted to limit my focus to the logic that handled parsing the `EDITOR` and finding it. I'll skip all the intermediate steps and experimentation to what I settled on:

1. More modular functions - easier to reason through and test.
2. Use an existing crate to replace the `which`/`where` logic, hence [which](https://crates.io/crates/which).
3. Use another existing crate to handle parsing `EDITORs` with arguments passed to them (e.g. `code --wait` or `vim -u NONE`), hence [shell-words](https://crates.io/crates/shell-words). Don't need advanced functionality, so don't go for the full, heavy-weight command-line parsers.
4. Simplify backup editor to only one per OS (can be changed back).
5. Make error handler nicer for PowerShell specifically.

Number 5 is sort of superfluous and a weird middle ground between just standardizing on outputting `EDITOR` for everything or giving specific error messages for POSIX and non-POSIX shells alike, like fish and nushell. However, my justification is 1) laziness, and 2) that PowerShell users on Windows (if you're using pwsh on Linux, you probably know what you're doing) are more likely to have trouble if the error isn't more explicit.

Linux/Mac OS users who use bash or zsh will have an error that works regardless of their technical level, and users who swap their default shell to something non-POSIX can be expected to know what to do with the error.

### Unit tests

The pre-existing 2 test failures on Windows are still there (#247), but there are no new test failures. There are still 3 ignored tests on Linux. I have also added 10 unit tests for `editor.rs`, which all succeed. The Linux output is much the same as the Windows output below:

```rust
> cargo test editor --package kanban-tui
...
     Running unittests src\lib.rs (target\debug\deps\kanban_tui-e894c5abd1da7d85.exe)

running 10 tests
test editor::tests::editor_env_hint_non_powershell ... ok
test editor::tests::parse_editor_handles_single_word ... ok
test editor::tests::parse_editor_handles_double_quoted_paths ... ok
test editor::tests::parse_editor_splits_program_and_args ... ok
test editor::tests::editor_env_hint_powershell ... ok
test editor::tests::parse_editor_handles_single_quoted_paths ... ok
test editor::tests::resolve_editor_falls_back_when_env_missing ... ok
test editor::tests::resolve_editor_handles_nonexistent_program ... ok
test editor::tests::resolve_editor_uses_env_var ... ok
test editor::tests::resolve_editor_handles_multiword_editor ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 158 filtered out; finished in 0.04s
...
```

There was a weird case after a fresh PC restart where the below test failed, but I have not been able to recreate the issue. If you think the test is malformed, let me know! I don't know if it could be due to the ordering of the tests for some reason, as they run in arbitrary order every time. But I structured the tests to fully set up and tear down their "environment" within each function, so I'm not sure how the error occurred.

```rust
running 10 tests
test editor::tests::editor_env_hint_powershell ... ok
test editor::tests::parse_editor_handles_single_word ... ok
test editor::tests::parse_editor_handles_single_quoted_paths ... ok
test editor::tests::parse_editor_splits_program_and_args ... ok
test editor::tests::parse_editor_handles_double_quoted_paths ... ok
test editor::tests::editor_env_hint_non_powershell ... FAILED
test editor::tests::resolve_editor_falls_back_when_env_missing ... ok
test editor::tests::resolve_editor_handles_nonexistent_program ... ok
test editor::tests::resolve_editor_uses_env_var ... ok
test editor::tests::resolve_editor_handles_multiword_editor ... ok

failures:

---- editor::tests::editor_env_hint_non_powershell stdout ----

thread 'editor::tests::editor_env_hint_non_powershell' (16172) panicked at crates\kanban-tui\src\editor.rs:116:9:
assertion `left == right` failed
  left: "$env:EDITOR"
 right: "$EDITOR"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Manual testing

(In addition to ensuring compliance with the other requirements in [CONTRIBUTING](https://github.com/fulsomenko/kanban/blob/develop/CONTRIBUTING.md))

I tested for:

- No keypress duplication
- Holding down a key repeats it
- Editing a field persists the change when the editor closes
- You can save the same temporary file multiple times while making edits
- You can edit a field that has been edited before

I manually confirmed that editing works for the following combinations:

#### Windows

- [x] Windows Terminal, PowerShell, `$env:EDITOR = "vim.exe"`
- [x] Windows Terminal, PowerShell, `$env:EDITOR` unset
  - Opens in `notepad`.
  - **NB:** Must save and close notepad tab to properly return to kanban. Document this behavior somewhere?
- [x] Windows Terminal, Nushell, `$env.EDITOR = "vim.exe"`
  - Opens in `vim`.
  - TIL how you set [environment variables](https://www.nushell.sh/book/environment.html#setting-environment-variables) in nushell. Turns out it has a different syntax than POSIX-like shells. But, I expect nushell users to be more knowledgeable than vanilla Windows users, so I'm not going through the effort of making the editor error statement prettier for a more niche shell. I would use an `enum` if so (e.g. pwsh, nushell, fish, posix). I think the simpler design of PowerShell/non-PowerShell is sufficient.
- [x] Windows Terminal, Nushell, `EDITOR` unset
  - Opens in `notepad`.
- [x] Git Bash, bash, `EDITOR` unset
  - Opens in `notepad`.
  - Previous behavior was to open in `vim` due to changes described above under "What".
  - Note: `vi` is also aliased to `vim` in my Git Bash.
  - No more recompile steps like in #247, almost certainly thanks to #257.
  -  On closing notepad window, Git Bash's window is not automatically reselected. I think this behavior also might have affected the other `notepad` test cases. I'm not going back to re-check though. This is fallback behavior, but it still makes it more annoying to switch between editing and browsing in the TUI, as you have to perform an extra action to select the screen.
- [x] Git Bash, bash, `export EDITOR='vim'`
  - Opens in `vim`.
  - No recompile steps triggered.

#### Linux

- [x] Windows Terminal, WSL bash, `EDITOR` unset
  - Opens in `vim` due to `vi` being aliased to that.
  - Full recompile triggered on `cargo run`, which is weird but not too weird. I had not previously run `cargo build` in WSL, so it's not totally surprising.
- [x] Windows Terminal, WSL bash, `export EDITOR='vim'`
  - Opens in `vim`.
  - Full recompile triggered due to changing above environment variable.
- [x] SSH over Windows Terminal to Debian 13, bash, `export EDITOR='vim'`
  - Opens in `vim`.
  - Started with this due to vim already being configured as my editor.
- [x] SSH over Windows Terminal to Debian 13, bash, `EDITOR` unset
  - Opens in `vim` due to `vi` being aliased to that.
  - No recompile steps triggered.

## Potential paths for fixing VS Code

I originally had more to say about my experimentation with VS Code. However, this is already quite long and isn't a priority for getting the PR submitted. Moreover, some of the open shell sessions and other in-memory data got lost when I restarted my PC, so I don't have all the steps and insights that I originally found. Those can be derived again from further experimentation. There's more I feel I'm missing, but I will leave this section bare bones for now.

Some potential inspiration to try to find a solution to VS Code's terrible file handling:

- [GitUI](https://github.com/gitui-org/gitui/blob/8619c07f3f6abc5df96ee2900c0107579b517f6d/src/popups/externaleditor.rs#L48)
- [Zellij](https://github.com/zellij-org/zellij/blob/0f6cf0664e05feab497778fbdddbf80492c252f2/zellij-server/src/os_input_output.rs)
- Check above repos for VS Code-related issues, closed or open

### Failed experimentation

- Using the absolute path for the temp file (by calling `canonicalize()` on `temp_file` before binding `status` in `edit_in_external_editor()`) breaks temp file behavior for vim, so that's a dead end.
- Adding another `disable_raw_mode()?` after the `execute!()` and before `flush`ing allows VS Code to open a file only once to attempt to edit a singular field. Saving and closing doesn't persist changes, and you get the same "file not found" error if you attempt to edit any field (same or otherwise) again. It's also way too hacky.

## Potential follow-up UX improvements

If you're stupid like me and forget how to unset a variable in bash (`unset VAR`), and instead do something like `export EDITOR=''`, then the runtime error for a missing editor will trigger. I find that the pop-up that shows the nice windowed error message disappears too quickly. It only lasts for one or two seconds. It would be nice to extend this at some point, but that is out of scope for this PR. (Pop-up not shown in screenshot below due to going away too quickly.)

image

## Weird behavior with file handling in latest develop

I discovered some behavior that I need to test thoroughly again and document. It is strictly speaking unrelated to the PR, but I found it while working on this, so I will add a comment unless you want me to create an issue. I don't think I'll try to track down the error myself and fix it myself (sorry!), in part because I'm not sure if it even is an error or some weird behavior. Regardless, I will be including info on that below when I have time.

For now, I declare this PR ready for review!


fulsomenko:
Nice! When time allows I will read through, test and review. Good job!

fulsomenko:
The changing of the vector of editors to only have the fallback on `vi` and `notepad` is a good resolution and clean up.

I don't think we have to be so concerned about wrecking someones workflow as a missing `EDITOR` would have led to the fallback vector either way. This is a welcomed change of behavior. Simple fallback and environment variable gives easy reasoning.

`patch` or `minor`. The next release will be a minor release regardless of this pr changeset. 
This as the changesets reduce the version bump down into `major > minor > patch`. For marking that this changeset brings a changed api, minor is a useful indicator.

still reading..

amanda-amy-frost:
Just added some edits to the README, which I specify towards the top of the edited comment above. I won't be touching it again until you're done :p Busy with other things anyway.

crates/kanban-tui/src/editor.rs
```diff
@@ -93,3 +97,97 @@ pub fn edit_in_external_editor(
 
     Ok(result)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::env::{remove_var, set_var};
```
fulsomenko: ### Must fix before merge — parallel test race

**Motivation.** `set_var`/`remove_var` mutate the process-global environment table. Rust's test runner executes all unit tests in parallel by default. When `editor_env_hint_powershell` sets `PSModulePath` at the same moment `editor_env_hint_non_powershell` removes it, one test reads the other's mutation — the thread that loses the race fails.

**Evidence.** The PR description already documents this exact failure:

```
thread 'editor::tests::editor_env_hint_non_powershell' panicked at editor.rs:116:
assertion `left == right` failed
  left: "$env:EDITOR"
 right: "$EDITOR"
```

That is not a fluke — it is this race made visible. The same race applies to all five `resolve_editor_*` tests that set/remove `EDITOR`.

**Fix — make both functions injectable so tests never touch global state.**

```rust
// editor_env_hint: take a bool, since the function only cares about presence.
fn editor_env_hint(is_powershell: bool) -> &'static str {
    if is_powershell { "$env:EDITOR" } else { "$EDITOR" }
}

// resolve_editor: keep the public 0-arg version, extract a testable inner fn.
fn resolve_editor() -> (PathBuf, Vec<String>) {
    resolve_editor_with(env::var("EDITOR").ok().as_deref())
}

fn resolve_editor_with(editor_env: Option<&str>) -> (PathBuf, Vec<String>) {
    let (program, args) = match editor_env {
        Some(value) => parse_editor(value),
        None => fallback_editor(),
    };
    let path = which::which(&program).unwrap_or_else(|_| PathBuf::from(program));
    (path, args)
}
```

Call site (currently line 73):

```rust
editor_env_hint(env::var_os("PSModulePath").is_some())
```

Tests become pure — no `set_var`, no `remove_var`, no race:

```rust
#[test]
fn editor_env_hint_powershell() {
    assert_eq!(editor_env_hint(true), "$env:EDITOR");
}

#[test]
fn editor_env_hint_non_powershell() {
    assert_eq!(editor_env_hint(false), "$EDITOR");
}

#[test]
fn resolve_editor_uses_env_var() {
    let (path, args) = resolve_editor_with(Some("vim"));
    assert!(path.ends_with("vim") || path.ends_with("vim.exe"));
    assert!(args.is_empty());
}

#[test]
fn resolve_editor_handles_multiword_editor() {
    let (path, args) = resolve_editor_with(Some("code --wait"));
    assert!(path.ends_with("code") || path.ends_with("code.exe") || path.ends_with("code.cmd"));
    assert_eq!(args, vec!["--wait"]);
}

#[test]
fn resolve_editor_falls_back_when_env_missing() {
    let (path, args) = resolve_editor_with(None);
    if cfg!(target_os = "windows") {
        assert!(path.ends_with("notepad") || path.ends_with("notepad.exe"));
    } else {
        assert!(path.ends_with("vi"));
    }
    assert!(args.is_empty());
}

#[test]
fn resolve_editor_handles_nonexistent_program() {
    let (path, args) = resolve_editor_with(Some("vi_and_emacs --flag"));
    assert_eq!(path, PathBuf::from("vi_and_emacs"));
    assert_eq!(args, vec!["--flag"]);
}
```

Total production-code delta: one `bool` param on `editor_env_hint`, one 2-line wrapper `resolve_editor`, one extracted `resolve_editor_with`. The `use std::env::{remove_var, set_var};` import in the test module goes away entirely.

**Minimal alternative** (if you'd rather not restructure): add `serial_test = "3"` as a dev-dep and put `#[serial(env)]` on the seven env-mutating tests. Works, but leaves global state in the tests. The injection above is preferred.

crates/kanban-tui/src/editor.rs
```diff
@@ -4,48 +4,51 @@ use crossterm::{
...
-        "notepad".to_string()
+        ("notepad".to_string(), vec![])
     } else {
-        "vi".to_string()
+        ("vi".to_string(), vec![])
     }
 }
 
+fn parse_editor(full_command: &str) -> (String, Vec<String>) {
+    let parts = shell_words::split(full_command).unwrap_or_default();
```
fulsomenko: ### Deferred — tracked as card #442 (added via #266)

**Motivation.** `unwrap_or_default()` silently swallows the `Err` case from `shell_words::split`.

**Evidence.** An unclosed quote (e.g. `EDITOR='vim`) returns `Err(MismatchedQuotes)`. After `unwrap_or_default()` the parts vec is empty, `program` becomes `""`, and the failure only surfaces later as a cryptic OS error from `Command::new("")` rather than an actionable message about the malformed `EDITOR` value.

Not blocking — fix detail on card **#442** (see #266).

crates/kanban-tui/src/editor.rs
```diff
@@ -93,3 +97,97 @@ pub fn edit_in_external_editor(
...
+
+    #[test]
+    fn parse_editor_splits_program_and_args() {
+        let (program, args) = parse_editor("vim -u NONE");
+        assert_eq!(program, "vim");
+        assert_eq!(args, vec!["-u", "NONE"]);
+    }
+
+    #[test]
+    fn parse_editor_handles_double_quoted_paths() {
```
fulsomenko, line 138: ### Deferred — tracked as card #443 (added via #266)

**Motivation.** Both quoted-path tests in this PR use forward slashes only. `shell-words` follows POSIX quoting rules where `\` is an escape character.

**Evidence.** A Windows user who sets `EDITOR=C:\Program Files\vim.exe` (unquoted, backslashes — a perfectly natural value to set on Windows) will have their path silently mangled: each `\` is consumed as an escape, producing something like `C:Program Filesvim.exe`. The current test suite would not catch this.

Not blocking — fix detail on card **#443** (see #266).

crates/kanban-tui/src/editor.rs
```diff
@@ -93,3 +97,97 @@ pub fn edit_in_external_editor(
...
+        assert_eq!(program, "C:/Program Files/VS Code/code.cmd");
+        assert_eq!(args, vec!["--wait"]);
+    }
+
+    #[test]
+    fn resolve_editor_uses_env_var() {
+        set_var("EDITOR", "vim");
+        let (path, args) = resolve_editor();
+
+        assert!(path.ends_with("vim") || path.ends_with("vim.exe"));
```
fulsomenko: ### Deferred — tracked as card #444 (added via #266)

**Motivation.** This assertion passes vacuously when `vim` is not installed.

**Evidence.** `which::which` falls back to `PathBuf::from(&program)` on failure (line 41). `PathBuf::from("vim").ends_with("vim")` is true regardless of whether `vim` is actually on PATH — so the test does not prove that `which` resolved anything. On a CI runner without vim this test passes but verifies nothing about the resolution behavior. The same problem affects `resolve_editor_handles_multiword_editor` (line 163).

**Recommended approach.** Drop `vim`/`code` as test fixtures and use `cargo` (or `rustc`) — guaranteed on PATH whenever `cargo test` is running this test. Then `assert!(path.is_absolute())` becomes a non-vacuous check: it's only true if `which::which` actually resolved the binary. No cfg-gating, no GHA package install, no coupling to optional Windows tooling.

Not blocking — fix detail on card **#444** (see #266).

fulsomenko (changes requested):
## Overview

Solid refactor. The four-function decomposition (`parse_editor` / `resolve_editor` / `fallback_editor` / `editor_env_hint`) is well-scoped, `shell-words` is the right tool for POSIX quoting (quoted paths with spaces now work), and replacing the `which`/`where` shell-out with the `which` crate is a clear improvement. One thing needs to be fixed before merge; three follow-ups have been tracked as kanban cards.

---

## What's good

- Clean single responsibilities per function — each does one job
- `shell-words` correctly handles `"C:/Program Files/code.cmd" --wait` — the core goal of this PR
- `which::which` is cross-platform and eliminates a shell-out entirely
- Error message shows `program.display()` and the shell-appropriate variable syntax (`$env:EDITOR` vs `$EDITOR`) — a concrete UX improvement
- 10 tests covering the right surface area: quoting, splitting, fallback, nonexistent program

---

## Must fix before merge

See the inline comment on line 104 — the `set_var`/`remove_var` calls across seven parallel tests cause a non-deterministic race. This is the cause of the intermittent failure described in the PR description. Full drop-in replacement is in that comment.

---

## Deferred — kanban cards created

Three issues are tracked for follow-up and do not block merge. The inline comments below pin each issue to its line; the fix detail lives on the card.

| Card | Location | Issue |
|------|----------|-------|
| **#442** | `parse_editor` line 29 | `unwrap_or_default()` silently swallows malformed `EDITOR` (unclosed quotes → empty program → cryptic OS error) |
| **#443** | `parse_editor` test line 134 | `shell-words` is POSIX-only; `C:\path\vim.exe` (unquoted backslashes) is mangled silently on Windows |
| **#444** | `resolve_editor_uses_env_var` line 152 | `ends_with("vim")` passes vacuously — `PathBuf::from("vim").ends_with("vim")` is true even when `which` fails |

crates/kanban-tui/src/editor.rs
fulsomenko: Verified empirically. Built a tiny crate against `shell-words = "1.1.1"` and ran:

```rust
for input in [
    r"C:\Program Files\vim.exe",
    r"C:\Users\foo\nvim.exe",
    r"C:\tools\code.exe --wait",
    r#""C:\Program Files\vim.exe""#,
    r#""C:\Program Files\vim.exe" --flag"#,
] {
    println!("{:?} -> {:?}", input, shell_words::split(input));
}
```

Results:

| `EDITOR=` value | `shell_words::split` result |
|---|---|
| `C:\Program Files\vim.exe` | `["C:Program", "Filesvim.exe"]` — two tokens, path mangled |
| `C:\Users\foo\nvim.exe` | `["C:Usersfoonvim.exe"]` — entirely nonsense path |
| `C:\tools\code.exe --wait` | `["C:toolscode.exe", "--wait"]` — path mangled, args OK |
| `"C:\Program Files\vim.exe"` | `["C:\\Program Files\\vim.exe"]` — quoted works |
| `"C:\Program Files\vim.exe" --flag` | `["C:\\Program Files\\vim.exe", "--flag"]` — quoted + arg works |

**Why this matters beyond "set EDITOR to a bare binary on PATH":** before this PR, `EDITOR` flowed straight into `Command::new(&editor)`, so a Windows user could set `EDITOR=C:\Users\foo\vim.exe` (the natural cmd / PowerShell syntax) and the OS path APIs would happily run it as a literal path — no PATH lookup needed. After this PR, the same string is fed through `shell_words::split` first, the backslashes are eaten as escapes, and `which::which("C:Usersfoovim.exe")` then `Command::new("C:Usersfoovim.exe")` both fail because that path doesn't exist anywhere.

So this isn't only an edge case for users with spaces in the path — anyone on Windows whose `EDITOR` is an absolute path (rather than a binary name expected on `PATH`) is affected. Workaround is to either quote the path or use forward slashes; worth documenting that explicitly in the VS Code limitations note in the README.

Details captured on card #443 (#266).

crates/kanban-tui/src/editor.rs
amanda-amy-frost: This is one of the things I forgot to comment on, because it was hard to keep track of everything. I hadn't tested thoroughly enough, but my expectation was that a malformed string would be fine because the runtime error message about the editor not being found would be sufficient for the user to determine what the cause was.

(**EDIT:** Which, as you make clear, that isn't the case. The failure mode is a lot less user-friendly.)

I was a bit iffy on `unwrap_or_default()`, but would `unwrap_or_else(full_command)` (and replacing the other default unwraps throughout the file in a similar manner) be sufficient to have the error bubble up nicely?

crates/kanban-tui/src/editor.rs
amanda-amy-frost: Would an unconditional find-replace for `\` to `/` before the `shell_words::split()` call be sufficient, or is that too simple and there could be some edge-case behavior? For example, could you first reduce the number of consecutive slashes to one before a find-replace, and then `split()`? Or, would a slightly more powerful crate do the job? I know there were some other ones I came across, and maybe one of them could be a drop-in replacement?

amanda-amy-frost:
I'll keep it brief for my last comment of the night. What I would like to do before merging is:

1. Address the testing race condition as described - thank you for the detailed feedback on this!
2. Find a better alternative to `shell_words`. Alternatively, exhaustively address all the non-POSIX cases. I prefer the first option.
3. Try `unwrap_or_else` instead of `unwrap_or_default` and see if the resulting behavior is acceptable.

I don't have a good solution out the gate for the vacuous passing tests, which is why I didn't address them as even though I was aware of the issue. (Again, hard to comment on everything when there's so much!) One lazy fix could be to install e.g. vim on the GitHub Actions runner, but that is unwieldy and creates a dependency/coupling that shouldn't exist and is vulnerable to change. (And it should be as stateless as possible.) Either way, I will leave that one up to you.

Hope you don't mind I attempt to address the above 3 things instead of just fixing the race condition alone. If you prefer, I can just do that and you merge the PR. Otherwise, I want to see the parts through that I think I can address 😊

Let me know either way. And thank you for the really good feedback and commentary!

As for that weird/failing file issue I'm having (I can't describe it well succinctly), would you mind if I create an issue for it that describes the behavior I've run into? I don't know if it should be a blocker for 0.5.0 if I were in your shoes, but it is something I want to describe tomorrow and make you aware of. It's related to the changes from #249, in that it breaks my ability to launch `kanban` with a file I've just created, at least on Windows.

**EDIT:** So I don't forget - also address crate versioning. Might ask you about that if necessary.

fulsomenko:
Heads up — addressed the Windows-only "Path traversal not allowed" failure you flagged in this thread. Tracked as KAN-445 and merged in #267.

## Root cause

`crates/kanban-service/src/path.rs::validate_path_with_cwd` was using `Path::canonicalize`, which on Windows always returns a verbatim UNC long-path (`\\?\C:\…`). `std::env::current_dir()` returns the plain form (`C:\…`). `Path::starts_with` is component-based, so the extra `\\?\` component made the in-cwd prefix check fail and the traversal guard rejected the path — only when the file already existed, because the non-existent-file branch took the lexical-fallback path and never canonicalized.

## Fix

Swapped `std::fs::canonicalize` for `dunce::canonicalize` on both branches and canonicalized `cwd` too, so the comparison is robust to symlinked or UNC-shaped cwds. Added two tests covering the previously-uncovered "file exists" case, plus a `test-windows` CI matrix entry so this class of bug is caught next time.

## References

Seemingly this is a known quirk and I looked to the `dunce::canonicalize` path for answers. I also wonder if this could help our other path woes.

- [rust-lang/rust#42869](https://github.com/rust-lang/rust/issues/42869) — the canonical upstream issue: "std::fs::canonicalize returns UNC paths on Windows"
- [rust-lang/rust#86447](https://github.com/rust-lang/rust/pull/86447) — Windows: fix fs::canonicalize to work with legacy drivers
- [`dunce` crate](https://crates.io/crates/dunce) — drop-in replacement for `std::fs::canonicalize` that strips `\\?\` when safe
- [`dunce::canonicalize` API](https://docs.rs/dunce/latest/dunce/fn.canonicalize.html)

Related real-world cases of the same bug shape:

- [vercel-labs/agent-browser#393](https://github.com/vercel-labs/agent-browser/issues/393) — Rust-built daemon returns UNC paths that crash Node.js
- [Mozilla bug 1581248](https://bugzilla.mozilla.org/show_bug.cgi?id=1581248) — `mach vendor rust` fails on Windows after Rust 1.37.0 with "Cannot mix UNC and non-UNC paths"
- [denoland/deno#12438](https://github.com/denoland/deno/pull/12438) — Deno had to add verbatim UNC prefix support
- [jj-vcs/jj#3986](https://github.com/jj-vcs/jj/issues/3986) — Jujutsu: "UNC paths do not work well with other programs"

#267 introduced the conflict you see here. Added the dunce dependency.

crates/kanban-tui/src/editor.rs
amanda-amy-frost: Changed and tested! Thanks for the details on this :p

crates/kanban-tui/src/editor.rs
amanda-amy-frost: Fixed and tested for some simple cases. See new tests. I haven't verified whether it is sufficient to handle more complex scenarios, but I'm first going to move on to finding a replacement for `shell_words`.

crates/kanban-tui/src/editor.rs
amanda-amy-frost: After some research, I found [typed_path](https://crates.io/crates/typed-path). That seems the most likely path (fuck, pun sort of intended) to try to handle cross-platform parsing. I don't have the time or energy to try to implement a proper solution, at least not today, so I will either leave it to you (or me in a separate PR) to fix :p

crates/kanban-tui/src/editor.rs
amanda-amy-frost: Fixed in the latest commit using `cargo` and `cargo version` for no args and args respectively. Thanks for the suggestion, makes my job super easy!

crates/kanban-tui/src/editor.rs
amanda-amy-frost: Also, this is how the runtime error looks now with `$env:EDITOR = "'vim.exe"` 😊

image

amanda-amy-frost:
DRAFTING

- Still check README
- `minor` preferred for this, or keep as `patch`? -> if Windows pathing was fixed, I would go minor myself, but there's already been a minor and patch release just in the last 24 hours
- Cards 442 and 444 fully resolved, 443 has potential pathway to fix (see my comment in the review)
- See latest X commits since your last comment

fulsomenko:
Yeah we go minor. Motivation being the changing behavior of the editor resolution. Lesson learned to keep the bump type of a pr self contained.

I wanted to push the release out yesterday as I had a deadline set to try and make [nixpkgs yarara release.](https://github.com/NixOS/nixpkgs/issues/503391) So I pushed 0.5.0.

For the completion of this adventure. I am happy to merge and close this pr once you feel happy with the results. Anything that can be deferred to a follow up can be amended later. 443..

Will review once more on your command

amanda-amy-frost:
Agh, I was hoping to delete that comment and replace it with the below before you replied to it! I haven't been working on this continuously the last hour or two, so that's why progress stalled for a bit.

I'll write the important details as quickly as I can:

- Failing `Test` check in CI is not due to me, so I will be ignoring :p (looks like you or some process fixed that already given the latest run?)
- I would say all the commits I've made today are ready for review! Besides those from today, do take a look at the README changes in particular and make sure they're to your satisfaction, because they might be different from what you first read given the chaos of yesterday.
- Card 443 will be left up to you. I have a lot of life things that will get in the way of me addressing that before you likely get around to it. Like I mentioned in the first draft of the first comment in the PR, I'm not likely to submit (m)any new PRs unless and until I run into more problems with my workflow 😅 Which is sort of a good thing or a bad thing, depending on how you look at it 😊
- Amazing work on the Windows file handling fix!! I tested it and it works beautifully.
- Note that all tests that run now pass on Windows, but the below 3 tests are ignored.
- I haven't read the docs or used `kanban` thoroughly enough, but the manual testing workflow has me often navigating back to a previous panel (or whatever you call it), meaning I hit `ESC` a lot. I would rather use `b` to go back and `ESC` for closing menus or quitting. (I've sometimes pressed `q` when I meant to hit `ESC` to go back and quit out of the whole thing.) I'm not going to describe it super thoroughly here, and obviously there are a lot of UX and consistency considerations, but last I checked - several days ago - there was something in the README (?) about allowing custom key configuration. Perhaps on the roadmap. No, I'm not going to read it right now and check 😤 Either way, I wanted to maybe ask about that: either changing default "back" behavior from `ESC` to `b`, or how config should work on Windows in particular, given my vague recollection of some config stuff causing testing errors until very recently.
- I will be doing final tests, but I would say everything is ready for review on your end, and if we both give the go ahead, then it's ready to be merged 😊 Whatever comments or questions I have related to the above or otherwise, I will ask in a separate comment after testing. Wanna finish that first before going on tangents like the above.

**EDIT:** The failing CI check in case you need to follow up on it:

https://github.com/fulsomenko/kanban/actions/runs/25863746793

**EDIT 2:** The ignored `cargo test` tests:

```rust
   Doc-tests kanban_persistence

running 2 tests
test crates\kanban-persistence\src\test_helpers\mod.rs - test_helpers::store_contract_tests (line 32) ... ignored
test crates\kanban-persistence\src\watch\file_watcher.rs - watch::file_watcher::FileWatcher (line 28) ... ignored
...
   Doc-tests kanban_tui

running 1 test
test crates\kanban-tui\src\state\mod.rs - state::SaveCoordinator (line 29) ... ignored
```

amanda-amy-frost:
Sorry for keeping you waiting! I had to pause in the middle of working on this as well.

## Noteworthy manual testing remarks

### On Windows

- Windows Terminal, PowerShell, `$env:EDITOR` unset
  - Full keyboard workflow possible with `Ctrl+S` and `Ctrl+W`, as closing the window in notepad selects WT again and allows for fluid keyboard control without additional input.
- Windows Terminal, nushell, `$env.EDITOR` unset
  - Notepad also works more cleanly as described above.
- Git Bash, bash, `$EDITOR` unset
  - Opens in notepad, but with that annoying behavior where notepad opens in the system tray and must first be selected. After closing notepad, Git Bash's window must reselected before keyboard input will be accepted. Given that the behavior is limited to a terminal that very few people are likely to use (certainly among those who use `kanban`), this is not an issue worth worrying about in the least.

### On Linux

- Windows Terminal, WSL bash, `EDITOR` set to `vim` or unset
  - *Full* recompile triggered on `cargo run` when switching from Windows. Without changing the `EDITOR` (as I did above in the first comment), switching immediately to Windows and doing `cargo run` will not trigger a recompile, but it will *always* compile on WSL before launching. I think I have something that could explain at least part of the data. See below.
- UNIT TESTING
  - The same 3 doc tests as shown above for Windows are ignored on WSL and Linux.

## WSL constant recompile

```rust
fn fallback_editor() -> (String, Vec<String>) {
    if cfg!(target_os = "windows") {
        ("notepad".to_string(), vec![])
    } else {
        ("vi".to_string(), vec![])
    }
}
```

I had mistakenly thought that `cfg!(target_os = "windows")` was a runtime check, but this check is still run at compile-time (?) as far as I can tell? Either way, this may in part play a role in triggering a recompile, but my ignorance is too vast on this issue. If that was the case, I would expect only some of the crates/modules to be recompiled, not the full 300-ish crates. Perhaps there is some underlying trigger that cannot be avoided.

Either way, it is immaterial for the time being. The fact remains that switching between Windows and WSL triggers a recompile in an unavoidable way. There must be some fallback editor for Windows and non-Windows alike, and while there are pure runtime options out there (presumably - I haven't tested them), it is unlikely to be the root cause, at least this small piece. Separate binaries should be recommended for Windows and WSL, so I have added that to the changeset and README.

## Cross-platform question

Do you use/test on a Mac? I ask because I was hoping to add "cross-platform" as a feature, which I had in an earlier commit for the README, before removing that in a later commit due to the Windows pathing issue. When that gets resolved, and unless I run into any other problems, I would definitely advertise that `kanban` works well across Linux, Mac, Windows, and WSL.

## Anything more?

I might be forgetting something. I have for example edited the changeset and README, so be sure to check those and let me know if they should be changed again in any way. I think any other notes or questions I have can wait until the PR is merged. I say it's ready 😊 And thanks and all the other niceties for making this process as smooth as possible!

fulsomenko:
Not to worry about keeping me waiting. I am not being blocked here.

My goal with `EDITOR` was to primarily to enable editing without leaving the terminal and secondly to not reinvent a text editor.

Yes. The cfg branching is compile-time macro that evaluates to a bool,  the branch is still compiled into the binary, just with the condition known at compile time, so the optimizer eliminates the dead branch.

If you want to fully exclude code from compilation (not just optimize it away), we can use the #[cfg(...)] attribute. The item will be removed entirely before the compiler even sees it. 

The recompilation you are facing between wsl and windows sounds very upsetting. I will give this some thought for resolution and see if I can find some time (and a windows box!) to implement.

I use and test on a Mac yes. My preferred OS is the one with freedom.
Some months back I actively developed with the darwin machine for this app. But nowadays mostly as a user.

Thank you for excellent contribution and dedicated teamwork.

fulsomenko (commented):
Thanks for the iteration on this. The must-fix race condition from the prior round is resolved cleanly, and the three deferred items are well captured on cards #442/#443/#444. A few notes from a fresh pass.

### Verifying the prior must-fix

The race in `editor_env_hint_powershell` / `editor_env_hint_non_powershell` is gone. Both `editor_env_hint` and `resolve_editor` now take their dependency as a parameter (`bool` and `Option<&str>` respectively), the tests are pure, and the `use std::env::{remove_var, set_var}` import has dropped out of the test module entirely. Exactly the shape the prior review suggested.

Card #442 (`unwrap_or_default()` swallowing `Err`) is addressed in 99979370: on a malformed `EDITOR` the match returns `(full_command.to_string(), vec![])` so `Command::new` surfaces the bad value in its error string. Pragmatic. Not as clean as a `Result` API would be, but fine for now and the followup card captures the cleaner redesign.

Card #444 (vacuous `vim`/`code` tests) is addressed in 30ed0e7e: tests use `cargo` and assert `path.is_absolute()`, which only holds if `which::which` actually resolved the binary. Now a real check.

### Blocker before merge

**Changeset TODO will ship into release notes.** The last bullet in `.changeset/editor-improve-robustness.md` still starts with `[TODO: REMOVE IF RESOLVED BEFORE RELEASE]`. Once the `changesets` tool collects this file, that line renders verbatim into the next release notes. Either resolve the Windows path issue (card #443) before tagging, or strip the marker and reword the bullet so the generated changelog is clean.

### Small follow-ups

1. **README typo.** \"swtiching\" should be \"switching\" in the new EDITOR configuration section.
2. **Missing semicolons in two tests.** `assert!(args.is_empty())` on `parse_editor_handles_malformed_single_word_command` and `parse_editor_handles_malformed_multiword_command` compiles fine but is inconsistent with the rest of the module.

### Code correctness

- The `parse_editor` fallback on `Err` is the right call for now. Small edge case worth noting: `parse_editor(\"\")` returns `(\"\", vec![])` via the `Ok(empty)` branch rather than the malformed branch. Matches prior behavior so not a regression, just unusual.
- Card #443 (Windows backslash mangling) persists by design and is captured in the changeset note. No action needed in this PR.

### Test coverage

10 tests, race-free, `which` resolution non-vacuously verified. One small gap: the wiring at the call site (`env::var_os(\"PSModulePath\").is_some()`) is not exercised. Trivial check, low priority.

### Risk

Dropping the `nvim` / `vim` / `nano` / `vi`-or-`notepad` fallback cycle is a real behavior change for users with broken or unset `EDITOR`. The description owns this tradeoff and the new behavior is more predictable. Low blast radius.

Performance and security are fine. `which::which` is one PATH scan per editor invocation, and `Command::new(&program).args(&args)` avoids shell injection.

### Verdict

Fix the changeset bullet, the README typo, and the two semicolons and this is ready from my side. Thanks again for picking this up and seeing it through the back and forth, and especially for the cross-platform manual testing matrix you ran. That kind of legwork is what makes review easy.

amanda-amy-frost:
Thanks for all this!

It's a small price to pay re: Windows/WSL. The fact alone that it works at all on Windows makes it that much easier for others to get into. And as long as the release binary isn't the 20-30 GiB that I have `cargo clean`ed about a dozen times now, I don't think it will present a huge barrier to entry :p

Regarding point 4 from the blockers before merging (before your edit), I would definitely add a note about scoping commit messages to crates (or however it should be done) [here](https://github.com/fulsomenko/kanban/blob/develop/CONTRIBUTING.md#commit-messages) in the CONTRIBUTING guide. I've kept my last commit to just `fix` to stay consistent with my style through now, and since the squash commit can handle the actual merge to `develop`.

Good luck with further development!

fulsomenko:
Oh that cargo weight is painful space and time. Have had some amount of worry that the long install and space taken will make anyone shy away. The cost is too big to pick the project up and iterate.

I like the commit style and scoping, will for sure put it in `CONTRIBUTING.md` when time given and wording is found. 
Until then, I am not going to let commit scope and message style block quality work. As you said, the merge squash will resolve it anywho. The changeset is for now the story teller of the pull.

Thank you for your effort, time and for contributing.
My eyes have now been set to open the door for windows support.

fulsomenko (approved):

amanda-amy-frost:
Just wanted to send you something while I remembered, since I got a notif for being assigned to this. I've been able to use 0.6.0 for a bit and have been really loving it! I haven't had time to use it enough to fully evaluate it, but I'm onboarded let's say. I just wanted to send you something after that point, as I am motivated myself to hear feedback from people who use what I make, and I was hoping to give you a boost out of the blue. I was going to comment on the 0.6.0 release itself but again haven't had time yet.

If you do want to hear some of the onboarding pains - the little quirks that might be worth fixing down the line for a smoother experience - let me know :) It's working flawlessly otherwise and I love the workflow! And thanks for your continual work on this!

PR #257, opened 2026-05-11
KAN-426 fix(tui): make KeyEventKind::Press filter unconditional across all platforms

fulsomenko:
Remove the `#[cfg(target_os = "windows")]` gates from the `KeyEventKind::Press` filter in `EventHandler`:

- Merge `KeyEventKind` into the existing crossterm import on line 1, removing the `#[cfg]`-gated separate import
- Remove the `#[cfg(target_os = "windows")]` attribute from the filter block, making it unconditional

**What**: The key event filter that drops non-`Press` events (Release, Repeat) now runs on every platform, not only Windows.

**Why**: The `#[cfg]` gate was added in #247 to fix key-doubling on Windows. However, the filter is a no-op on Linux/macOS in standard terminal mode — crossterm only emits `KeyEventKind::Press` there unless the kitty keyboard protocol (`PushKeyboardEnhancementFlags`) is active, which this project does not use. The guard created unnecessary platform divergence and made the filter untestable on Linux/macOS.

**How**: Two-line change in `crates/kanban-tui/src/events.rs`: fold `KeyEventKind` into the existing import and drop the `#[cfg]` attribute from the `if` guard. No logic added or removed.

**Testing**: `cargo test --workspace` (no regressions), `cargo clippy --all-targets --all-features -- -D warnings` (clean, no unused-import warnings).

### DIFF
--- a/.changeset/kan-426-replace-cfg-gated-keyeventkind-filter.md
+++ b/.changeset/kan-426-replace-cfg-gated-keyeventkind-filter.md
@@ -1,0 +1,9 @@
+---
+bump: patch
+---
+
+Make KeyEventKind::Press filter unconditional across all platforms (KAN-426)
+
+- The Windows-only `#[cfg(target_os = "windows")]` gate on the key event filter is removed — the filter now runs on all platforms
+- On Linux/macOS crossterm only emits `Press` events in standard terminal mode, so the filter is a no-op there; behaviour is unchanged on all platforms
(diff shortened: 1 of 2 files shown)


amanda-amy-frost:
LGTM 😊 Thank you for doing it!

**EDIT**: And to be clear, I think the universal approach is best. It was suggested as a runtime check in #247, but this is better for sure!

PR #298, opened 2026-05-25
fix(tui): refresh task list before selecting moved card

fulsomenko:
## Summary
- Fixes selector not following card after h/l move in ColumnView and GroupedByColumn views
- Root cause: select_card_by_id ran against a stale CardList cache before prepare_frame refreshed it
- Fix mirrors the KAN-403 approach already applied to create_card: call prepare_frame() before select_card_by_id() at both move handler call sites

## Test plan
- [x] test_move_card_right_selects_moved_card_in_kanban_view passes
- [x] test_move_selected_cards_right_selects_first_moved_card passes
- [x] Existing stale_model_regression_tests stay green

### DIFF
--- a/crates/kanban-tui/src/handlers/card_handlers.rs
+++ b/crates/kanban-tui/src/handlers/card_handlers.rs
@@ -472,6 +472,7 @@ impl App {
                 }
             }
 
+            self.prepare_frame();
             self.select_card_by_id(card_id);
         }
     }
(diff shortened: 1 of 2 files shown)


fulsomenko:
Replaced by https://github.com/fulsomenko/kanban/pull/303 (branch renamed to KAN-437/fix-tui-selector-does-not-follow-card-after-move-across-columns-in-kanban-view)

PR #299, opened 2026-05-25
fix(persistence-sqlite): validate required string fields in SQLite write path

fulsomenko:
## Summary
- Adds a `required_str` helper that returns a `KanbanError::Serialization` on empty strings
- Applies it to `SprintLog.status` in `write_card_with_conn`, closing the silent empty-string insertion gap
- Also applies it to `board.name`, `column.name`, `card.title`, and `board.sprint_names` items, which all feed NOT NULL columns with no DEFAULT or CHECK constraint

## Context
The original `upserts.rs` was removed in KAN-348. The remaining gap was in `sqlite_store.rs` where `SprintLog.status: String` bypassed the `NOT NULL` enforcement because SQLite only rejects NULL, not empty strings. An empty `status` would insert silently and produce an unloadable row when later deserialized back into a domain type.

## Test plan
- [x] `test_empty_sprint_log_status_returns_error` passes (was failing before the fix)
- [x] `cargo test --package kanban-persistence-sqlite` passes (28 tests, 0 failures)
- [x] `cargo test --workspace --exclude kanban-tui` passes (all suites green; the 2 TUI failures are pre-existing red tests from a separate in-progress card)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes (0 warnings)
- [x] `cargo fmt --package kanban-persistence-sqlite --check` passes


fulsomenko:
Replaced by https://github.com/fulsomenko/kanban/pull/304 (branch renamed to KAN-250/replace-unwrap-or-default-with-required-str-in-sqlite-upserts)

PR #301, opened 2026-05-25
KAN-174 fix(tui): retain card selection after toggle-completion in kanban view

fulsomenko:
Toggling a card's completion status in the kanban (column) view now keeps the card selected after the operation:

- Add `prepare_frame()` before `select_card_by_id` in `toggle_card_completion` and `toggle_selected_cards_completion` — matching the KAN-403 fix already applied in `create_card`
- Extend `select_card_by_id` to search all column lists when the card is not in the active one, navigate to the containing column, then select — making the method robust for any future cross-column operation
- Add two regression tests in `stale_model_regression_tests.rs` covering single-card and multi-select toggle paths in kanban (column) view

**What**: When toggling a card's completion status, the service layer auto-moves the card to the Done column. Without a view refresh before the selection update, the next render frame clears the old column list and the selector drops silently.

**Why**: `toggle_card_completion` and `toggle_selected_cards_completion` called `select_card_by_id` on a stale view (same root cause as KAN-403 in `create_card`). The kanban column view keeps a separate `CardList` per column; after the card moves, the active column list no longer contains it so `select_card` returns false and the selection is lost on the next `prepare_frame()` call.

**How**: Two-part fix. First, `prepare_frame()` is called before `select_card_by_id` in both toggle paths so the column lists are current when the selection runs. Second, `select_card_by_id` itself is made defensive: if `select_card` fails on the active list it searches all task lists via `get_all_task_lists()`, resolves the column index containing the card, calls `try_set_active_column_index()` to navigate there, and then selects.

**Testing**: `cargo test -p kanban-tui --test stale_model_regression_tests` (12 pass including 2 new), `cargo test --workspace` (no regressions), `cargo clippy --all-targets --all-features -- -D warnings` (clean).

## Test plan

- [x] test_toggle_card_completion_retains_selection passes
- [x] test_toggle_multi_card_completion_retains_selection passes
- [x] Existing stale_model_regression_tests stay green
- [x] cargo fmt --check passes

### DIFF
--- a/.changeset/kan-174-toggle-completion-tests.md
+++ b/.changeset/kan-174-toggle-completion-tests.md
@@ -1,0 +1,13 @@
+---
+bump: patch
+---
+
+In the kanban (column) view, toggling a card's completion status now keeps the card
+selected after the toggle.  Previously, the card would be moved to the Done column
+by the service layer, but the selection would silently drop on the next render frame
+because the view was not refreshed before the selection was updated.
+
+`select_card_by_id` has also been made robust for any view: if the card is not found
+in the currently active column list it now searches all column lists, navigates to the
+column that holds the card, and selects it there.  This prevents silent selection drops
+whenever a card moves between columns as a side effect of an operation.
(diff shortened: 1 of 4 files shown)


fulsomenko:
Replaced by https://github.com/fulsomenko/kanban/pull/307 (branch renamed to KAN-551/fix-tui-retain-card-selection-after-toggle-completion-in-kanban-view)

PR #302, opened 2026-05-25
feat(tui): surface save errors to user via UI banner

fulsomenko:
## Summary
- Adds save_error field to App and an error channel between the save worker and the event loop
- Save failures (disk full, permissions, etc.) are forwarded to App state instead of only being logged
- Renders a persistent red banner with warning icon above the footer; clears on next successful save

## Test plan
- [x] Tests covering save error state on App pass (5 new tests in save_error_tests.rs)
- [x] cargo fmt --check passes


fulsomenko (commented):
Self-review: The diff is clean. The save_error_rx channel is properly created in spawn_save_worker and stored in PersistenceState. The select! branch follows the same pattern as save_completion_rx. The UI renders a 1-row red bar between the main view and the footer, which persists until the next successful save. No must-fix issues found.

fulsomenko:
Replaced by https://github.com/fulsomenko/kanban/pull/306 (branch renamed to KAN-174/add-user-visible-save-error-notifications)

PR #340, opened 2026-06-13
KAN-706 refactor(persistence): add suppress_next_write default no-op to ChangeDetector

fulsomenko:
Add a default no-op `suppress_next_write` method to the `ChangeDetector` trait in `kanban-persistence`:

- Add `fn suppress_next_write(&self) {}` default method to `ChangeDetector` in `crates/kanban-persistence/src/traits.rs`
- Add `test_suppress_next_write_default_noop_does_not_require_override` verifying structs that do not override the method compile and call it without effect
- Add `test_file_watcher_satisfies_change_detector_with_new_method` verifying `FileWatcher` satisfies the trait and dispatches through the vtable

**What**: `ChangeDetector` gains a fifth method with a default empty body. Existing implementors require no changes.

**Why**: The upcoming `WebSocketChangeDetector` (for the HTTP collaborative backend) implements `ChangeDetector` but does not need write suppression -- the server tags each push event with a `writer_instance_id` so clients can filter their own writes. Adding the default no-op here lets `WebSocketChangeDetector` skip suppression while `FileWatcher` will override the method with its counter logic in the next card (KAN-707). This follows the same pattern as `PersistenceStore::close()`.

**How**: Trait default method with empty body and a doc comment explaining the contract. `FileWatcher::suppress_next_event` (the existing concrete method) is not touched in this card -- that rename is KAN-707.

**Testing**: `cargo test --package kanban-persistence` (32 tests pass, 2 new), `cargo clippy --all-targets --all-features -- -D warnings` (clean).


fulsomenko:
Closing: the restructuring moves suppression into JsonDataStore.save() directly rather than on the ChangeDetector trait. ChangeDetector stays at four methods. suppress_next_write as a trait method is no longer needed. KAN-710 becomes the load-bearing card for this change.

PR #345, opened 2026-06-14
KAN-729 feat(domain): wrap commands in CommandBatch with correlation_id, issued_by, and AppType

fulsomenko:
Introduce `CommandBatch` as the audit-log entry type: one `execute()` call is one atomic transaction recorded as one batch carrying shared provenance. Stored in full, never stripped.

## What

- `CommandBatch { commands: Vec<Command>, correlation_id, issued_by, app_type, app_version, session_id, timestamp }` in `kanban-domain` — the audit-log entry shape.
- `AppType` enum in `kanban-core` (`Tui | Cli | Mcp | HttpClient | Unknown`, `#[non_exhaustive]`) for command attribution.
- `CommandStore` trait is now batch-oriented: `append_batch` / `batch_count` / `load_batches` / `load_all_batches`, all returning full `CommandBatch` values.
- The in-memory store holds `Vec<CommandBatch>` intact — no strip-on-write. The audit log *is* the batch.
- `KanbanContext` carries a `session_id` (minted once per process at open) and an `app_type` (via `with_app_type`). `execute()` builds exactly one `CommandBatch` per transaction.
- All three surfaces self-attribute: TUI (`AppType::Tui`), CLI (`AppType::Cli`), MCP (`AppType::Mcp`).
- `Command` and all inner command structs derive `PartialEq`.

## Why

Every mutation needs provenance for the upcoming HTTP collaborative backend: who issued it, from which app, in which session, with a correlation id for tracing. The audit log must be able to answer "who changed what, when, from where."

The type started as a per-command `CommandEnvelope`. Designing the normalized audit-log persistence (KAN-474) revealed that provenance is **batch-level**: one `execute()` is one atomic transaction (commits or rolls back together via `with_transaction`) and one HTTP request (`POST /v1/commands`). The commands in it are not independently retryable, so they share one provenance — not N. Per-command envelopes would repeat the same metadata N times on the wire and would not map to the header/detail tables. So the type was reshaped to batch-level.

## How

- One `execute()` -> one `CommandBatch::new(commands, correlation_id, issued_by, timestamp, app_type, app_version, session_id)` -> one `append_batch`.
- The audit log stays **dumb**: append-only, opaque, never read back for correctness (undo uses the in-session `UndoStack`, not the log). No idempotency / dedup constraint lives here — that is a server-request-layer concern, tracked as KAN-751, deliberately not built until the server exists.
- Normalized on-disk persistence (SQLite header/detail tables) is deferred to KAN-474; this PR keeps the in-memory log and establishes the type + trait + attribution.

## Testing

`cargo test --workspace` (2137 tests pass), `cargo clippy --all-targets --all-features -- -D warnings` (clean), `cargo fmt --all --check` (clean). New coverage: batch construction/defaults, multi-command preservation, serde round-trip, full-provenance storage (no strip), batch-boundary preservation, and an end-to-end `execute -> load_all_batches` test asserting populated provenance.


fulsomenko:
Closing to reopen with corrected branch and changeset naming (KAN-729/* convention).

PR #371, opened 2026-07-12
fix(domain): deterministic board ordering (fixes #358)

### LINKED ISSUE #358: Projects random order
The order of the Projects are random, every time I open kanban the order change randomly.

kanban 0.7.2

fulsomenko:
## Problem
Fixes #358 — the Projects list reorders randomly on every launch.

## Root cause
`snapshot_impl` sorted boards with `sort_by_key(|b| b.position)`. Boards created before the `position` field existed all deserialize with the default `position = 0`. When every board shares the same key, the stable sort's tiebreaker is the source order — and boards are collected from a `HashMap`, whose iteration order is non-deterministic across runs. The snapshot (and thus the persisted + displayed order) therefore reshuffles each launch.

## Fix
Sort boards by `(position, created_at, id)` so equal positions resolve deterministically by creation time (and `id` as a final tiebreak). No data migration required.

## Tests
- Added `test_snapshot_orders_boards_with_equal_position_by_created_at`.
- Full `kanban-domain` suite passes (572).

### DIFF
--- a/crates/kanban-domain/src/in_memory_store/snapshot.rs
+++ b/crates/kanban-domain/src/in_memory_store/snapshot.rs
@@ -6,7 +6,12 @@ impl InMemoryStore {
         let state = self.read_state()?;
 
         let mut boards: Vec<_> = state.boards.values().cloned().collect();
-        boards.sort_by_key(|b| b.position);
+        boards.sort_by(|a, b| {
+            a.position
+                .cmp(&b.position)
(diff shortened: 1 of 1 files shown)


fulsomenko:
Superseded by #374 — re-targeted onto card-tracked branch KAN-848/fix-domain-deterministic-board-ordering-fixes-358.

PR #372, opened 2026-07-12
fix(tui): set task priority with `p` from the card list (fixes #360)

### LINKED ISSUE #360: Tasks priority change from list
In the Tasks list there is the shortcut suggestion `p Set task priority` but it does nothing.

kanban 0.7.2

fulsomenko:
## Problem
Fixes #360 — the task list shows the hint `p Set task priority`, but pressing `p` does nothing.

## Root cause
Two bugs:
1. The `p` keybinding in `card_list.rs` pointed at `KeybindingAction::EditCard` instead of a priority action.
2. `app/input_router.rs` had no `KeyCode::Char('p')` arm in Normal mode (only `P` for bulk), so `p` fell through to the no-op catch-all.

## Fix
- New `handle_set_card_priority_key`: activates the highlighted card and opens the single-card `SetCardPriority` dialog with the card's current priority preselected — mirroring the existing `CardListAction::TogglePriority` flow.
- New `KeybindingAction::SetCardPriority`, mapped in `execute_action` and repointed from the `card_list` binding.
- Router arm for lowercase `p`, gated to the cards panel.

## Tests
- `tests/card_priority_key_tests.rs`: `p` on a highlighted card opens the dialog with the right preselection; `p` on the projects panel is inert.

### DIFF
--- a/crates/kanban-tui/src/app/input_router.rs
+++ b/crates/kanban-tui/src/app/input_router.rs
@@ -201,6 +201,12 @@ impl App {
                     self.pending_key = None;
                     self.handle_toggle_task_list_view();
                 }
+                KeyCode::Char('p') => {
+                    self.pending_key = None;
+                    if self.focus.active == Focus::Cards {
+                        self.handle_set_card_priority_key();
+                    }
+                }
                 KeyCode::Char('P') => {
                     self.pending_key = None;
                     self.handle_set_selected_cards_priority();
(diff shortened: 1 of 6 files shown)


fulsomenko:
Superseded by #375 — re-targeted onto card-tracked branch KAN-849/fix-tui-set-task-priority-with-p-from-the-card-list-fixes-360.

PR #390, opened 2026-07-19
KAN-862 feat(service): gate content mutations on archived boards — read-only shelf (C10-Foundation)

fulsomenko:
## What & why

An archived board is a **read-only shelf** (DECIDED on the SE-C epic): it can be listed, viewed, restored, or permanently deleted, but its contents cannot be edited in place — to change it, restore it first. This is the enforcement (write-gate) half of the C10 read-only-shelf model; C10a (#388) was the read half. Without it, an archived board's content was silently mutable by UUID (the surfaces aren't shipped yet, so real exposure is ~0 — this lands the guard before they do).

Spiked in a throwaway worktree through both backends before implementing; that spike also uncovered and got the SQLite restore data-loss bug fixed first (KAN-863, #389).

## Approach — one seam, fail-closed

Enforced once at the command-execution seam (`KanbanContext::execute`) so every surface (TUI/CLI/MCP) inherits identical behavior — no per-operation scatter. Before the transaction, a guard resolves each command's target board(s) and rejects the batch with a new `BoardArchived` error if any is archived.

- The classifier is an **exhaustive match** over every `Command` variant — a new variant won't compile until it declares its class (fail-closed). Lifecycle (archive/restore/delete-board) and internal/synthetic commands resolve to no gated board and pass through.
- A `Card` is column-scoped, so `card -> board` hops through the column.
- New `DomainError::BoardArchived` (+ `is_board_archived()` predicate, `ErrorCode::BOARD_ARCHIVED`, HTTP 409, client-safe message).
- `create_card`/`create_sprint` resolve the owning board from the archived collection too, so a create on an archived board reaches the guard (`BoardArchived`) instead of failing early with a misleading `NotFound`.

## Tests (TDD, red first)

`crates/kanban-service/tests/archived_board_write_gate.rs` — content mutations (create card/sprint, move card, edit column, edit board settings) rejected with `BoardArchived`; lifecycle ops, live-board mutations, and post-restore edits allowed. Covered **in-memory and through the real SQLite backend** (per the new all-backend CLAUDE.md rule). Full kanban-service + kanban-domain suites green; clippy `-D warnings` + fmt clean; CLI/MCP/TUI build clean.

## Follow-ups

The read-only shelf is now enforced service-side. The surface slices (C10b CLI, C10c MCP, C10d TUI) expose the archived-board views on top; C7/C8/C6a add the archive lifecycle surfaces.

### DIFF
--- a/crates/kanban-domain/src/error.rs
+++ b/crates/kanban-domain/src/error.rs
@@ -140,6 +140,12 @@ pub enum DomainError {
     #[error("column {column_id} has reached its WIP limit of {limit}")]
     WipLimitExceeded { column_id: Uuid, limit: u32 },
 
+    /// Returned when a content mutation targets an archived board (or an entity
+    /// belonging to one). Archived boards are a read-only shelf; the client must
+    /// restore the board before editing its contents.
+    #[error("board {board_id} is archived; restore it before editing its contents")]
+    BoardArchived { board_id: Uuid },
+
     #[error(
         "sprint {sprint_id} belongs to board {sprint_board} but card is being created on board {card_board}"
     )]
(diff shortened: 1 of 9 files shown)


fulsomenko:
Closing unmerged: the read-only-shelf model this enforces was reversed in design discussion. Direction is now a UNIFIED archival-as-marker model — an archived board/card is the same fully-functional entity with a reference in the archival table; the live/archived distinction lives only at the interfaces (CLI/TUI/MCP). Storage already implements the marker model for both boards (board_archival) and cards (archived_cards is a pure reference with a guarded delete_card). Superseded by a Phase-0 design pass on the unified model. The create_card/create_sprint archived-aware resolution here is salvageable for that work (in git history).

PR #411, opened 2026-07-20
KAN-911 feat(tui): archived boards reuse the full live-board UI 1:1

fulsomenko:
## Context

Follow-up to KAN-891 under root KAN-864 ("an archived board is just a board; the full existing board UI should work 1:1"). KAN-891 let you OPEN an archived board and see its tasks, but did so with a parallel `active_archived_board_index` and routed only ~5 of ~89 board-resolution sites through an archival-aware `viewed_board()`, setting `active_board_index = None` when an archived board is active. So card detail, board settings, sprints, columns and most card operations were dead for archived boards, and the input router dispatched a drilled-in archived board through a tiny archived-only keyset.

## The accessor

- `App::viewed_board_id()` (from the existing `viewed_board()`) and `Model::board_by_id()` — the archival-agnostic key that resolves a live OR drilled-in archived board (checks both `boards()` and the archived flat list).
- `App::is_board_active()` distinguishes an activated/viewed board from one merely highlighted in the projects list (used by the empty-state render messages).

## Consumer categories routed through the accessor

- **Card operations** (`card_handlers`): create/move/toggle/assign-sprint/order/sort/hide-assigned/sprint-filter/manage-children gates and board resolution.
- **Card detail & sprint-detail actions** (`detail_view_handlers`): `a` assign-sprint, sprint-detail Assign/MoveColumn, extracted `handle_open_selected_board_sprint()`.
- **Board detail / settings / sprints / columns** (`detail_view_handlers`, `column_handlers`, `sprint_handlers`): settings edit, branch-prefix, sprint nav counts, sprint activate/complete/create, column create/rename/delete/move — all now resolve the viewed board, not the live-list highlight.
- **Filters, popups, dialogs** (`filter_handlers`, `popup_handlers`, `dialog_handlers`, `selection_dialog`, `filter_popup`): sprint pickers and task-sort commands.
- **Render / UI** (`render_strategy`, `ui/card_detail`, `ui/sprint_detail`, `ui/dialogs/*`): task-list panels, card/sprint detail, create-card and view-select dialogs.
- **Input router**: when drilled into an archived board, keys dispatch through the full Normal handler while `mode` stays `ArchivedBoardsView` so the projects panel and its navigation bounds remain archived-scoped (KAN-893). `handle_edit_board_key` opens BoardDetail for a drilled-in archived board.

## Kept on the live list (not archival-agnostic)

Projects-panel selection bounds, activate/deactivate, delete-board index shifting, rename/export board — these legitimately mean "the live board LIST", so archived heads never leak into it.

## Tests

Parity tests drive the SAME handlers/key events a live board uses on a drilled-in archived board seeded with a column, cards and a sprint:
- `test_archived_board_card_detail_matches_live`
- `test_archived_board_settings_view_reachable`
- `test_archived_board_sprints_view_reachable`
- `test_archived_board_card_action_works` (toggle completion via the shared handler)
- `test_projects_panel_still_lists_live_boards_only` (GUARD: the live projects list excludes archived boards)

All kanban-tui tests green (293 lib + integration), `clippy -D warnings` clean, `fmt` applied; `kanban-cli`/`kanban-mcp` compile.

## Not 1:1 (minor)

`q` while drilled into an archived board is a no-op rather than quitting (Esc returns to the archived list, which is tested and the intended exit); the archived-view `q` on the list still toggles back to live.

### DIFF
--- a/crates/kanban-tui/src/app/card_selection.rs
+++ b/crates/kanban-tui/src/app/card_selection.rs
@@ -117,10 +117,9 @@ impl App {
 
     pub fn populate_sprint_task_lists(&mut self, sprint_id: uuid::Uuid) {
         let cards = self.model.cards();
-        let board_opt = self
-            .selection
-            .active_board_index
-            .and_then(|i| self.model.boards().get(i));
+        // Archival-agnostic: resolve the viewed board (live or drilled-in
+        // archived) so an archived board's sprint task lists populate 1:1.
+        let board_opt = self.viewed_board();
 
         let (uncompleted_ids, completed_ids) = if let Some(board) = board_opt {
             let columns = self.model.columns();
--- a/crates/kanban-tui/src/app/clipboard.rs
+++ b/crates/kanban-tui/src/app/clipboard.rs
@@ -9,22 +9,19 @@ impl App {
         F: Fn(&Card, &Board, &[Sprint], &str) -> String,
     {
         if let Some(active_id) = self.selection.active_card_id {
-            if let Some(board_idx) = self.selection.active_board_index {
-                let boards = self.model.boards();
-                if let Some(board) = boards.get(board_idx) {
-                    if let Some(card) = self.model.card(active_id) {
-                        let sprints = self.model.sprints();
-                        let output = get_output(
-                            card,
-                            board,
-                            sprints,
-                            self.app_config.effective_default_card_prefix(),
-                        );
-                        if let Err(e) = clipboard::copy_to_clipboard(&output) {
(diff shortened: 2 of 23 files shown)


fulsomenko:
Superseded: this added MORE archival special-casing (active_archived_board_index, viewed_board dual-resolver, is_board_active, archived-only dispatch) rather than removing it. Per LSP an archived board must be substitutable for a live one everywhere; the archival distinction belongs only at consumption sites (the board list's live-vs-archived filter). KAN-911 recast to that target. Keeping the parity tests + model.board_by_id from this branch for the rework.

PR #426, opened 2026-07-20
KAN-936 fix(domain): make CompactColumnPositions archival-aware (compact over Include)

fulsomenko:
## What

KAN-936 (ARCH-DECOR C2), the follow-up the F1/C1 spike surfaced.

C1 made `move_card` place an archived card at a coherent ordinal over the Include set. But `CompactColumnPositions::execute` still renumbered the **live-only** set (`list_cards_by_column`), and the TUI runs it right after `ArchiveCards` (`app/animation_tick.rs`). So a live card could be renumbered onto the coherently-placed ordinal of an archived card (which stays live behind a marker and is untouched by archive), re-colliding it.

This switches the compaction read to the F1 filtered path with `ArchivedFilter::Include`, so every card in the column gets a distinct dense ordinal and each subset (live view, archived view) renders in correct relative order. `capture_inverse` now captures over the Include set too, matching `execute` (which renumbers archived cards as well).

## How the command reaches archived cards

`CompactColumnPositions::execute` runs in-memory against `CommandContext.store` (a `&dyn DataStore` backed by `InMemoryStore`). Archived cards stay in the `cards` map behind a marker in `StoreState.archived_cards`; `InMemoryStore` overrides `list_cards_by_column_filtered` (F1c) to honour `Include`, returning live + archived for the column. Switching the read from `list_cards_by_column` to `list_cards_by_column_filtered(column_id, Include)` is the whole fix.

## Tests (Red → Green, per backend)

Domain unit (in-memory):
- `test_compact_preserves_distinct_ordinals_across_live_and_archived`
- `test_archive_then_compact_keeps_archived_card_coherent`

Service contract (in-memory, JSON, SQLite via `context_contract_tests!`):
- `test_archive_then_compact_keeps_archived_ordinal_distinct` — archive the middle of three cards, compact, save+reload, assert no live/archived ordinal collision and the Include view is dense.

All went **red** on the old live-only impl (verified: live card collided at the archived card's position 1) and green after the fix, on all three backends.

## Verification

- `cargo test -p kanban-domain` (637 pass) and `cargo test -p kanban-service --features test-helpers` green
- `cargo clippy --all-targets -D warnings` clean on both crates
- `cargo fmt --check` clean

### DIFF
--- a/crates/kanban-domain/src/commands/card/positioning.rs
+++ b/crates/kanban-domain/src/commands/card/positioning.rs
@@ -55,7 +55,13 @@ pub struct CompactColumnPositions {
 
 impl CompactColumnPositions {
     pub fn execute(&self, context: &CommandContext) -> KanbanResult<()> {
-        let cards = context.store.list_cards_by_column(self.column_id)?;
+        // KAN-936 (ARCH-DECOR C2): compact over the Include set (live +
+        // archived). Archived cards stay live in `cards` behind a marker and
+        // keep their coherently-placed ordinal (C1); renumbering the live-only
+        // set would re-collide a live card onto an archived ordinal.
+        let cards = context
+            .store
+            .list_cards_by_column_filtered(self.column_id, crate::ArchivedFilter::Include)?;
         for (i, mut card) in cards.into_iter().enumerate() {
             if card.position != i as i32 {
                 card.position = i as i32;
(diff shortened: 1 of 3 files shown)


fulsomenko:
Closing: making CompactColumnPositions Include-aware breaks the established close-the-gap UX (test_multi_column_archive_compacts_every_affected_column expects the kept card to compact to position 0 after archiving). Decision: compaction stays live-only. Archived cards keep frozen positions; the archived view sorts by (position, created_at). C1's Include-move is retained for the archived-move-collision fix. See KAN-914 O1 refinement.

PR #455, opened 2026-07-22
fix(persistence): V8->V9 migration writes atomically (KAN-966)

fulsomenko:
## Bug (v0.8.0 bug-hunt, epic KAN-960)

The async `migrate_v8_to_v9` step wrote the upgraded envelope with a bare `tokio::fs::write` (truncate in place), while every other migration step and its own sync twin use `AtomicWriter::write_atomic` (temp file + rename). A crash mid-write could leave the live file truncated, recoverable only via the `.v8.backup`.

## Fix

Use `AtomicWriter::write_atomic`, closing the crash-truncation window. Adds an on-disk round-trip test that migrates a real V8 file and asserts a valid V9 result.

persistence-json suite green; clippy -D + fmt clean.

### DIFF
--- a/.changeset/kan-966-v8v9-atomic-write.md
+++ b/.changeset/kan-966-v8v9-atomic-write.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+---
+
+Hardened the JSON V8 to V9 migration to write atomically (temp file plus rename),
+matching every other migration step. Previously it wrote the upgraded file in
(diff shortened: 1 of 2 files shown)


fulsomenko:
Closing as stale/redundant: verified the fix (`v8_to_v9_archived_boards.rs` using `AtomicWriter::write_atomic`) is already present on `develop` via changeset `max-kan-966.md`, landed through a different route. This PR's diff would be a no-op.

PR #456, opened 2026-07-22
KAN-967 fix(mcp): set_board_sort echoes the resolved value, not null

fulsomenko:
## Bug (v0.8.0 bug-hunt, epic KAN-960)

`set_board_sort {"sort":"name"}` (order omitted) persisted a concrete order to disk but the JSON response reported `board_sort_order: null` because it echoed the raw request (`order.map(...)`) rather than the resolved value. The CLI reports the resolved pair, so the two surfaces disagreed.

## Fix

`McpContext::set_board_sort` now returns the resolved `(field, order)` pair (it already computed it to persist), and the tool echoes that. An omitted dimension is reported concretely.

## Test

MCP integration: `set_board_sort` with `order` omitted echoes a concrete `board_sort_order` string (was `null` before). mcp suite green; clippy -D + fmt clean.

### DIFF
--- a/.changeset/kan-967-set-board-sort-echo.md
+++ b/.changeset/kan-967-set-board-sort-echo.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+---
+
+Fixed the MCP `set_board_sort` tool reporting `null` for a dimension that was
+omitted from the request but actually resolved and persisted. When you set only
+the field (or only the order), the response now echoes the concrete value that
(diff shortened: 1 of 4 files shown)


fulsomenko:
Closing as stale/redundant: verified the fix (`set_board_sort` returning the resolved `(field, order)` tuple) is already present on `develop` via changeset `max-kan-967.md`, landed through a different route. This PR's diff would be a no-op.

PR #492, opened 2026-07-28
fix(persistence): detect own file writes by content, not event timing

fulsomenko:
## Summary
- `FileWatcher`'s own-write suppression used a fixed count (2 OS events) after arming, which under contended builds (e.g. Nix sandboxed test runs) can be exhausted by more raw events than expected, letting the real own-write event leak through as a false external-change notification (`test_own_write_decrements_suppress_counter` flake).
- Own writes are now identified definitively by comparing the `instance_id` already stamped into every JSON envelope's metadata against our own instance id, no timing or event-count guess involved. The old count-based suppression is kept only as a fallback for when ownership can't be determined from content (no instance id configured, non-JSON content), reimplemented as a real expiring window instead of a raw counter.
- Wires the backend's `instance_id()` into `FileWatcher` on startup so the TUI actually gets definitive detection.

## Test plan
- [x] `cargo test -p kanban-persistence --lib` (new content-based ownership tests, rewritten suppression-window tests)
- [x] `cargo test -p kanban-tui --lib`
- [x] `cargo clippy -p kanban-persistence -p kanban-tui --lib --tests -- -D warnings`
- [x] `cargo fmt`

### DIFF
--- a/crates/kanban-persistence/src/watch/file_watcher.rs
+++ b/crates/kanban-persistence/src/watch/file_watcher.rs
@@ -2,11 +2,53 @@ use crate::traits::{ChangeDetector, ChangeEvent};
 use crate::{PersistenceError, PersistenceResult};
 use chrono::Utc;
 use notify::{RecursiveMode, Watcher};
-use std::path::PathBuf;
-use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Arc;
+use serde::Deserialize;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, Mutex as StdMutex};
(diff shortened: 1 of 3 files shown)


fulsomenko:
Closed as a side effect of renaming the head branch (`fix/persistence-own-write-detection` → `max/kan-1006`) to match this repo's established branch-naming convention — GitHub auto-closes a PR when its head branch is deleted rather than retargeting it, which I hadn't accounted for. No commits were lost; they're preserved (plus a changeset commit that was missing from this PR) on `max/kan-1006`. Tracked as kanban card 1006. Replacement PR incoming.

PR #521, opened 2026-08-01
KAN-1039 feat(web): Topcoat web-client prototype with remote reads

fulsomenko:
Adds `kanban-web`, a server-rendered web-client prototype built on [Topcoat](https://github.com/tokio-rs/topcoat), and a real remote-read path in `kanban-backend-http` so it can point at a live `kanban-server` instead of only a local file.

- New `kanban-web` crate: a single `/` page rendering boards → columns → cards straight off `KanbanContext`, no REST/DTO layer in between
- `HttpBackend` (`kanban-backend-http`) gains real implementations of `get_board`/`list_boards`/`get_column`/`list_columns_by_board`/`list_cards_by_column` (previously all `unsupported()`), backed by real HTTP requests against `kanban-server`'s existing routes
- Fixes a real bug found along the way: `HttpBackend`'s sync-to-async bridge panics ("Cannot start a runtime from within a runtime") when called from an already-async caller, which `kanban-web`'s topcoat page handler is — fixed by running the page's DataStore reads through `tokio::task::spawn_blocking`
- `kanban-web` picks a local JSON file (`KANBAN_FILE`) or a remote server (`KANBAN_SERVER_URL`) at runtime
- Renamed the crate from the original spike name `kanban-web-topcoat` to `kanban-web`

**What**: A working (read-only) web client, plus enough of `HttpBackend` for it to optionally run against a remote server instead of a local file.

**Why**: Evaluating Topcoat as a server-rendered frontend for this project's REST API, decoupled from a hand-written DTO/API client layer. Tracked on the kanban board as KAN-1039 (the spike, with full execution learnings) and KAN-1040 (the `HttpBackend` read-path work, spawned from the spike).

**How**: TDD throughout — every capability (the Topcoat page, each `HttpBackend` read method, the remote-backend page render) has a red test committed before its green implementation. `HttpBackend`'s known, documented fidelity gaps: `Board`'s allocation counters (`card_counter`, `next_sprint_number`, `sprint_counters`, ...) have no exposing endpoint and are defaulted — safe only because every `HttpBackend` write path is still `unsupported()`; `get_card` stays `unsupported()` because no route returns a card's `board_id` without the caller already knowing it.

**Testing**: `cargo test -p kanban-web -p kanban-backend-http` (23 tests, including 9 new HTTP-backend integration tests against a real `kanban-server::test_helpers::TestServer`), `cargo clippy`/`cargo fmt` clean on both crates, `cargo build --workspace` succeeds, and verified live end-to-end: started a real `kanban-server`, seeded a board/column/card over its REST API with curl, pointed `kanban-web` at it via `KANBAN_SERVER_URL`, and confirmed the rendered HTML.

**Note for reviewers**: this branch also carries the `rust-overlay`/clippy-1.97 toolchain bump and its unrelated lint fixes (needed because Topcoat's MSRV is 1.95, higher than the previously-pinned toolchain) — split out as standalone PR #520 so it can be reviewed and merged independently. Once #520 merges, this branch should be rebased onto `develop` to drop the now-duplicate commits from this diff.

## Test plan
- [x] `cargo test -p kanban-web -p kanban-backend-http` (23 tests, all green)
- [x] `cargo clippy -p kanban-web -p kanban-backend-http --no-deps -- -D warnings`
- [x] `cargo fmt -p kanban-web -p kanban-backend-http -- --check`
- [x] `cargo build --workspace`
- [x] Live end-to-end: real `kanban-server` + `kanban-web` pointed at it via `KANBAN_SERVER_URL`, rendered the seeded board/column/card


amine2233:
Hi @fulsomenko, I'm waiting for this feature, do you know it will be available soon or not ?

PR #562, opened 2026-08-11
KAN-1110 refactor(backend): make with_transaction a required method and override it per backend

fulsomenko:
Completes the transaction work started by KAN-1067 and KAN-1071. `KanbanBackend::with_transaction` loses its generic default and becomes a required method, so every backend states its own strategy instead of inheriting one.

## Why the default had to go

```rust
fn with_transaction(&self, f: &mut dyn FnMut() -> KanbanResult<()>) -> KanbanResult<()> {
    let before = self.snapshot()?;
    match f() { ... self.apply_snapshot(before) ... }
}
```

Two problems. It calls `DataStore::snapshot`/`apply_snapshot`, which KAN-1079 removes, so it stops compiling regardless. And it gives a silently wrong answer for backends it does not suit: `HttpBackend` talks to a REST API and cannot roll back, but it inherited this default and failed through `snapshot()`, which returns unsupported. It failed by accident rather than by decision.

Deleting the default turns "you might not have thought about transactions" into a compile error.

## What each backend now says

- SQLite: native BEGIN/COMMIT/ROLLBACK (KAN-1067, already merged)
- JSON: single-lock snapshot and restore via the inherent methods (KAN-1071, already merged)
- InMemoryStore: snapshot and restore, which is genuinely cheap for a hashmap. Uses the `snapshot_impl`/`apply_snapshot_impl` pair KAN-1071 widened to `pub`
- HttpBackend: `unsupported("with_transaction")`, naming itself
- StubBackend and MockBackend: matching their siblings

## FnMut rather than FnOnce

This card rewrites every impl, so it is the cheapest point to settle a question KAN-1105's spike raised: moving an owned value into the closure needs an `Option::take()` dance, because a bare `move ||` will not compile as `FnMut` even though the trait calls it exactly once.

Kept `FnMut`. Calling an `FnOnce` requires ownership of the callee, so it cannot be called through `&mut dyn`. The alternative is `Box<dyn FnOnce()>`, which forces a heap allocation at every call site across five crates to avoid one working idiom in one place. The trait now documents the idiom instead.

## Tests

`test_http_backend_with_transaction_returns_unsupported` is genuinely red before this change. Its captured failure names the exact bug:

```
error must name with_transaction itself, not a helper it happens to call
through a fallback (got: "operation not supported by this backend: snapshot")
```

`test_in_memory_with_transaction_rolls_back_full_graph` (board, column, card, sprint, dependency edge) passes before and after, because the old default already handled in-memory correctly. It is regression cover, recorded as such rather than presented as red.

The rest of the red surface is the compile gate: `cargo test --workspace --no-run` fails for any implementor that does not supply an override. That gate passes here, which is what proves none was missed.

`cargo test --workspace --no-run` clean, backend crate tests and the 5 service `with_transaction` tests green, clippy `--all-targets --all-features -D warnings` clean across the four touched crates plus `kanban-server`, fmt clean.

### DIFF
--- a/.changeset/kan-1110.md
+++ b/.changeset/kan-1110.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+KanbanBackend::with_transaction loses its generic default and becomes a required method, so each backend states its own transaction strategy instead of inheriting one that suited only in-memory stores. HttpBackend now reports with_transaction as unsupported directly rather than failing through an inherited snapshot call, and InMemoryStore keeps the snapshot and restore approach that is cheap for a hashmap.
--- a/crates/kanban-backend-http/src/lib.rs
+++ b/crates/kanban-backend-http/src/lib.rs
@@ -24,6 +24,13 @@ impl kanban_backend::KanbanBackend for HttpBackend {
     fn instance_id(&self) -> uuid::Uuid {
         self.instance_id
     }
+
+    fn with_transaction(
+        &self,
+        _f: &mut dyn FnMut() -> kanban_domain::KanbanResult<()>,
+    ) -> kanban_domain::KanbanResult<()> {
+        Err(kanban_domain::KanbanError::unsupported("with_transaction"))
+    }
 }
 
 impl HttpBackend {
--- a/crates/kanban-backend-http/src/lib.rs
+++ b/crates/kanban-backend-http/src/lib.rs
@@ -133,4 +133,21 @@ mod tests {
         assert_eq!(backend_ref.instance_id(), backend.instance_id);
         Ok(())
     }
+
+    #[test]
+    fn test_http_backend_with_transaction_returns_unsupported() -> kanban_domain::KanbanResult<()> {
+        let backend = HttpBackend::new("http://example.com")?;
+        let backend_ref: &dyn kanban_backend::KanbanBackend = &backend;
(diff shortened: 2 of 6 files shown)


fulsomenko:
Closing in favour of #563, which covers the same card.

Both branches delete `with_transaction`'s generic default and add the four missing overrides. They diverge on the signature question the card asked to be settled here: this PR keeps `&mut dyn FnMut` and documents the `Option::take()` dance, #563 moves to a boxed `FnOnce` so callers can move owned values in directly. We went with `FnOnce`; it also let `undo()` and `redo()` drop the borrow aliases they were keeping purely to satisfy `FnMut`.

#563 additionally resolves the card's open race question (investigated, not reachable, every consumer already serialises), shares the combined rollback error between the snapshot-restoring backends, and covers the rollback-failure path on both of them.

Nothing here is lost. Both tests from this branch exist in #563 by the same names, and this branch's `test_in_memory_with_transaction_rolls_back_full_graph` was the better of the two: it asserts whole-snapshot equality rather than enumerated entities, and its batch adds as well as deletes so rollback is checked in both directions. Both improvements are carried across in 20d413a6.

One note for whoever picks up similar work: the doc here says `f` is called exactly once, but this branch's own `HttpBackend` override declines without calling it. #563 documents it as at most once.

PR #578, opened 2026-08-14
KAN-1176 feat(domain): keep column default_status in step with completion columns

fulsomenko:
Found by executing a later slice and hitting a failure, not by review.

The V14 and v8 migrations made every column's `default_status` agree with its board's `completion_column_ids`. That fixed data at rest. It did not fix the write path: `Board::update` sets the board's list and never touches any column, so the two sources diverge again the moment anyone configures completion columns at runtime.

That is not a legacy-data edge case. It is the only way boards are configured today. Repointing the lifecycle rules onto the derived helpers without closing this gap broke ~24 tests across 8 binaries.

Rules now applied whenever a board's completion columns change:

- a column added to the list gets `default_status = Done`
- a column removed from the list is reset to `Todo` only if it was `Done`
- a removed column carrying anything else is left alone, so this never silently overwrites a deliberate choice

Undo is handled by snapshotting each touched column's prior `default_status` before the change executes, and appending one column update per touched column to the inverse batch. Restoring the board's list without restoring the columns would leave the two sources diverged after an undo, which is the bug this closes.

`ApplyBoardSettings` also writes `completion_column_ids` directly and needed the same treatment. The card only named `UpdateBoard`; the second path came from re-running the enumeration rather than trusting it.

Nothing is removed here. `--completion-columns` and `completion_column_ids` both stay; later slices remove them. This card only makes the two sources agree.

Verified by applying the saved repoint patch on top of this branch and re-running every previously-failing binary: `completion_sync` 17/17 (was 11 failed), `archival_contract` 247/247, `move_cards` 26/26, `cli_integration` 162/162, `mcp integration` 96/96, `batch_completion_sync` 2/2, `stale_model_regression` 15/15.

### DIFF
--- a/.changeset/kan-1176-config-writes-default-status.md
+++ b/.changeset/kan-1176-config-writes-default-status.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+Keep column default_status in step with a board's completion-column configuration
--- a/crates/kanban-domain/src/commands/board_commands.rs
+++ b/crates/kanban-domain/src/commands/board_commands.rs
@@ -512,18 +512,41 @@ impl ApplyBoardSettings {
     /// `Editable::from_entity` impl, then re-apply that DTO via another
     /// `ApplyBoardSettings`. The DTO covers exactly the fields this command
     /// writes, so the round-trip is symmetric.
+    ///
+    /// Re-dispatching `ApplyBoardSettings` alone is not enough for
+    /// `completion_column_ids`: its forward `execute` re-syncs every touched
+    /// column's `default_status` against the DTO's list, so a column that had
+    /// a deliberate non-`Done` status before the original change would land on
+    /// `Todo` (the sync's default for a removed column) instead of its prior
+    /// value. Snapshot every touched column here, before `execute` runs, and
+    /// append an explicit restore per column so it wins over the re-sync.
     pub fn capture_inverse(&self, store: &dyn DataStore) -> KanbanResult<Vec<Command>> {
         let board = match store.get_board(self.board_id)? {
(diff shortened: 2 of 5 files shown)


fulsomenko:
Review found an undo-fidelity defect; fixed in 2d0bf63e and 10aba4a7.

`ApplyBoardSettings::capture_inverse` captured only the board DTO and relied on re-dispatching itself to re-derive column statuses on undo. That works when the prior value was `None` or `Done`, but loses a deliberate non-`Done` value:

1. a column sits at `InProgress`, outside the completion list
2. adding it to completion sets it to `Done`
3. undo re-dispatches with the prior list, sees the column as removed and currently `Done`, and resets it to `Todo`

The "leave a non-`Done` value alone" guard cannot help, because by undo time the column really is `Done`. Reproduced before fixing:

```
assertion `left == right` failed: undo must restore the deliberate InProgress, not overwrite it with Todo
  left: Some(Todo)
 right: Some(InProgress)
```

`ApplyBoardSettings::capture_inverse` now snapshots the touched columns before execute and appends an explicit restore per column, mirroring `UpdateBoard::capture_inverse`. Confirmed that `capture_inverse` does run before `execute`, so the snapshot reads genuinely prior state.

Worth noting all nine checks were green on the previous commit. Nothing in the gates exercises undo fidelity for a value the tests never seeded.

The writer enumeration was re-run rather than trusted. `DeleteColumn::capture_inverse` recreates the column carrying its own `default_status` verbatim rather than re-deriving it, so it has no equivalent gap. Every other path dispatches through `UpdateBoard`, and no creation path can set completion columns at all.

PR #590, opened 2026-08-16
KAN-1212 feat(persistence-sqlite): prefixes table, schema v9 to v10, additive backfill

fulsomenko:
Adds the `prefixes` table and backfills it from existing board and sprint prefixes. Additive: nothing reads the table yet, so this changes no behaviour. The allocation card wires it up.

## Why prefixes become an entity

Card identifiers collide today. Two boards created without an explicit prefix both inherit the configured default, so a card on each is `KAN-1`. Making the prefix a first-class row with a unique key makes an identifier globally unique, which is what lets a later card resolve one in a single indexed lookup instead of loading every card.

## Collision policy, and one the fixture found

A board whose prefix collides gets an integer-incremented derivation: `task, task2 … task42`. Uniquely-prefixed boards are untouched. Two boards *explicitly* set to `DUP`/`dup` abort the open naming the prefix rather than silently renaming — a user-chosen collision is not the same as a default-inherited one.

The real fixture surfaced a case neither the card nor its spike anticipated: **two boards both leaving `sprint_prefix` unset collide on the `"sprint"` default**, exactly as they would on `"task"`. The increment policy now covers the sprint-naming axis too.

## The merge rule

One row carries both counters only when a board uses the same string for `card_prefix` and `sprint_prefix` — true of the real tracker's main board. When they differ, two rows, each carrying only its own counter. When `sprint_prefix` is null, one row.

That rule is keyed on the board's own two columns, not on a sprint's `card_prefix` override; the card's prose conflated them. The differing case is the untested path in practice, since the real tracker only exercises the equal one, so it has a dedicated test using `board update --sprint-prefix`.

## Identifier preservation

`test_migrate_v9_to_v10_preserves_every_cards_dynamically_resolved_identifier` computes each card's identifier through the current dynamic chain (`sprint.card_prefix → board.card_prefix → "task"`) before the migration and asserts it is unchanged after.

That matters because a later decision stores a card's prefix so a prefix change stops renaming historic cards. A backfill that failed to reproduce today's resolution would break exactly the identifiers the decision exists to protect.

## A 42x migration speedup found during implementation

First open on a 204-board / 5007-card / 203-sprint fixture was **4.2s**, because 408 row inserts were each an implicit fsync-bound transaction. Wrapping them in one transaction brought it to **0.10s**. Second open is 0.018s.

## Scope

Table and backfill only. No `Card.prefix` column — that belongs to the allocation card, per this card's own note.

The migration writes only the transform; the version gate, the `VACUUM INTO` backup and the version bump stay in the central `open()` path. Raw column reads are guarded with `pragma_table_info`, because `migrate()` runs unconditionally against every earlier boundary's test fixtures, several of which predate the columns this step reads.

## Verification

Fixture generated by a real pre-change binary at `fd1e3ba9` and committed as a test asset, rather than hand-authored — hand-written old-format fixtures have twice passed while proving nothing on this project.

11 migration tests, 3707 workspace tests, clippy `-D warnings`, fmt, and the factory compile-lock all clean. Red and green are separate commits; the red run failed 9 of 11 against unmodified code.


fulsomenko:
Superseded by #592, which carries this table plus the JSON side and the shared-namespace model that replaces owner_kind/owner_id.

PR #591, opened 2026-08-16
KAN-1213 feat(persistence-json): envelope V14 to V15, additive prefixes backfill

fulsomenko:
The JSON counterpart to the SQLite prefixes table (#590). Bumps the envelope to V15 and backfills a `prefixes` collection from existing board and sprint prefixes. Additive: nothing reads it yet.

## A real-binary fixture caught a defect the hand-authored ones missed

The card mandates fixtures generated by a real pre-change binary. The first pass hand-authored them instead, and the tests passed.

Replacing them with a fixture built by an actual pre-V15 binary — carrying an explicit prefix, a board whose card and sprint prefixes differ, the default fallback, a genuine two-board collision, and a sprint-level override — immediately failed:

```
expected a row for rel, got: ["kan", "dev", "task", "task2", "auth"]
```

A board's sprint-naming prefix is a separate namespace from its card prefix. When they coincide one row carries both counters; when they differ the board owns two rows. The backfill read only `card_prefix` and hardcoded `sprint_counter` to zero, so `card_prefix=DEV, sprint_prefix=REL` produced no `rel` row at all.

**The SQLite backfill has always emitted both**, so the two backends silently disagreed about what a migrated store contains. Nothing reads the table yet, so no user is affected — but the counter-removal card assumes both backends carry the same data, and would have discovered this only after deleting the legacy source.

The existing JSON/SQLite agreement test did not catch it, because it compares against a recorded expectation rather than running both migrations, and no hand-authored fixture happened to set a differing `sprint_prefix`. That test's weakness is now documented rather than assumed away.

## What is in this card

- `FormatVersion::V15` and the `prefixes` collection in the envelope
- The backfill transform, projecting fields off the raw `Value` rather than deserializing — the migration path must read files written before the record existed
- Both migration chains updated. The chain is not a loop: `Migrator::migrate` hardcodes the destination in a match arm with the version baked into every string, so V15 meant rewriting that arm and extending `run_split_and_upgrade_chain` in both the async and sync paths
- Collision policy matching SQLite: `task, task2 …`, with uniquely-prefixed boards untouched

## Verification

- `test_migrate_v14_to_v15_on_a_real_binary_fixture` — the fixture above, with its provenance recorded in a doc comment
- `test_migrate_v14_to_v15_identifier_preservation` — computes each card's dynamically-resolved identifier before the migration and asserts it is unchanged after, which is what a later card's stored-prefix decision depends on
- `test_sync_and_async_chains_produce_identical_v15_prefixes_output` — the dedicated V15 equality test the first pass omitted
- `test_migrate_v14_to_v15_collision_resolution_matches_sqlite_v9_to_v10_sequence`

3707 workspace tests, clippy `-D warnings`, fmt clean. Red and green are separate commits; the red run failed all 8 against an `unimplemented!()` stub.

## One tidy-up included

`transform_v14_to_v15` was exposed as ordinary public API with no production caller. It is now behind `cfg(test)` or a `test-helpers` feature, with a self dev-dependency so the integration tier — a separate crate — can still reach it. Several migration tests also asserted version 15 while their failure messages still said V14.

## Known gap, not closed here

The JSON/SQLite agreement test compares against a recorded expectation rather than executing both migrations in one process. Closing that properly needs a harness spanning both backend crates, and it should land before the counter-removal card relies on the two agreeing.


fulsomenko:
Superseded by #592, which carries this migration plus the SQLite side and the shared-namespace model that replaces owner_kind/owner_id.

PR #600, opened 2026-08-16
KAN-1214 feat(domain,persistence-sqlite,persistence-json): allocate sprint numbers from the shared prefix row

fulsomenko:
Sprint numbering moves onto `prefixes.sprint_counter`, the same shared namespace the card axis already allocates from. Two boards sharing a sprint prefix now draw from one counter instead of each handing out number 1.

The old `ensure_sprint_counter_initialized` scan is deleted rather than ported. It seeded a counter from `MAX(sprint_number)` for one board, which is precisely what lets a shared namespace issue the same number twice. `board.sprint_counters` is still written behind the row, clamped upward only, matching how `CreateCard` maintains `board.card_counter`. KAN-1216 removes both.

## A counter-semantics defect this exposed

Both legacy counters hold the NEXT number to hand out, and `card_counter` starts at 1 on a brand-new board. A prefix row holds the LAST number used, so its next allocation is `counter + 1`. The V15 backfill copied both across verbatim, giving one struct two opposite meanings and making the first card and the first sprint allocated after migrating each skip a number.

Nothing read `Prefix.sprint_counter` yet, which is why no test caught it. The card half had already shipped in KAN-1247.

The projection now converts, clamped at zero. Verified against the real 1216-card tracker migrating V12 to V16:

| | highest existing | prefix row | first allocation after migrating |
|---|---|---|---|
| cards | KAN-1248 | 1248 | KAN-1249 |
| sprints | 22 | 22 | 23 |

Contiguous on both axes. Before this change the row read 1249 and 23, and the next allocations were KAN-1250 and sprint 24.

V15 is unreleased and no store has been migrated by it, so correcting the projection in place is sufficient. No corrective migration is needed.

## Test changes worth reading

Two tests in `resolve.rs` constructed cross-board sprint-number ambiguity by creating a sprint on each of two default-prefix boards. Those boards now share a namespace and get 1 and 2, so the tests give them distinct prefixes instead. The ambiguity they guard is still reachable, just not through the default prefix.

The remaining updated expectations are the semantic shift itself. The two invariant tests in `migration_v9_to_v10.rs` now compare next-to-issue on both sides rather than raw values, which is what they meant all along.

- `cargo test --workspace --all-features` green
- `cargo clippy --all-targets --all-features -- -D warnings` green
- `cargo fmt --all` applied

### DIFF
--- a/crates/kanban-backend-memory/src/in_memory_store/prefixes.rs
+++ b/crates/kanban-backend-memory/src/in_memory_store/prefixes.rs
@@ -18,14 +18,20 @@ impl InMemoryStore {
 
     pub(crate) fn upsert_prefix_impl(&self, prefix: Prefix) -> KanbanResult<()> {
         let mut state = self.write_state()?;
-        let normalized = Prefix::normalize(&prefix.name);
+        // Normalised on the way in rather than trusted from the caller, so the
+        // stored name satisfies `Prefix`'s invariant however it was built.
+        // SQLite gets this for free: its ON CONFLICT never updates the name.
+        let prefix = Prefix {
+            name: Prefix::normalize(&prefix.name),
+            ..prefix
+        };
         // Replace in place rather than push-and-dedup: two spellings of one
         // name are one namespace, and a second row would let two owners
         // allocate the same number.
         match state
             .prefixes
             .iter_mut()
-            .find(|p| Prefix::normalize(&p.name) == normalized)
+            .find(|p| Prefix::normalize(&p.name) == prefix.name)
         {
             Some(existing) => *existing = prefix,
             None => state.prefixes.push(prefix),
--- a/crates/kanban-domain/src/commands/mod.rs
+++ b/crates/kanban-domain/src/commands/mod.rs
@@ -1,5 +1,5 @@
 use crate::data_store::DataStore;
-use crate::{DomainError, KanbanError, KanbanResult};
+use crate::{KanbanError, KanbanResult};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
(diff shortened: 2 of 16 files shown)


fulsomenko:
Folded in two defects found reviewing the merged epic. Both were verified with a probe before being written up, and each has a red test committed ahead of its fix.

## A WIP-rejected create burned a card number

KAN-1247 moved allocation out of `CreateCard::execute` into the service, but the WIP check is the first statement *inside* `execute`. So the counter was bumped and persisted, then the command rejected, reserving a number for a card that never existed. Before that change the number came from `board.card_counter`, bumped after the check, so rejection cost nothing.

The rule is now `kanban_domain::wip::check_wip_limit` taking a `&dyn DataStore`, called from the service before it allocates. `CommandContext::check_wip_limit` delegates to it, so the command still checks and a replayed log cannot exceed a limit either. One rule, two tiers, the same shape `allocate_card_number` already uses across that boundary. Allocation could not move back into the command because its serialized shape is frozen around `card_number`.

Verified against a copy of the real tracker:

```
created: KAN-1251
Error: column 591afb61... has reached its WIP limit of 1
after rejection: KAN-1252
```

## upsert_prefix disagreed across backends about the stored name

SQLite's `ON CONFLICT` omits the name from its SET list, so the row keeps its normalised name. The in-memory path replaced the whole struct and took the caller's casing, and JSON inherits that. `Prefix::name` is documented as always normalised.

```
SQLITE rows: [Prefix { name: "kan", card_counter: 9, ... }]
MEMORY rows: [Prefix { name: "KAN", card_counter: 9, ... }]
```

Latent rather than user-visible, since both lookup paths normalise, but it is the same shape as the two-truths backfill divergence this epic already fixed. Normalised on the way in now.

The existing contract case asserted the row count and both counters but never the name, which is why it passed on both backends. It asserts the name now.

- `cargo test --workspace --all-features` green, 189 suites
- `cargo clippy --all-targets --all-features -- -D warnings` green
- `cargo fmt --all` applied

PR #602, opened 2026-08-17
KAN-1252 refactor(domain): Card::new takes a board id instead of &mut Board

fulsomenko:
`Card::new` took `&mut Board` and called `board.get_next_card_number()`, so every construction mutated allocation state. That blocked removing `Board.card_counter` (KAN-1216), because the signature has ~325 call sites.

The count overstated it. There was exactly **one** production caller, `CreateSubcardCommand`, and it did not want either value `Card::new` derived:

```rust
let (prefix, card_number) = allocate_card_number(...)?;   // allocates properly
let mut card = Card::new(&mut board, ...);                // bumped card_counter for nothing
card.card_number = card_number;                           // overwrote it
card.prefix = prefix;                                     // and the prefix
```

So the bump consumed a `card_counter` value on every subcard create and `upsert_board` persisted it. Harmless only because nothing reads that counter for allocation any more.

Everything else was test scaffolding. `Card::new` is now what it already was in practice — a plain test constructor taking `board_id`, with `card_number: 0` and `prefix: String::new()`. `Card::create` remains the production constructor.

## Pure refactor

The subcard `CreateSubcardCommand` produces keeps the same id, number, prefix, title, position and parent edge. The only non-test production edit is that call site.

## The hazard, and how it was handled

Some tests relied on `Card::new` handing out increasing numbers. With `card_number: 0` they all collapse to 0, so anything sorting by number or asserting uniqueness fails. Those failures were signal, not churn: **31 call sites now set `card_number` explicitly**, so each test states the numbering it actually depends on rather than inheriting it from a constructor side effect.

Exactly one test was deleted, `test_card_new_uses_board_card_counter_no_prefix_arg`, which asserted the removed behaviour. All 7 deleted assertions in the diff are inside it. No other assertion was weakened or removed.

- `cargo test --workspace --all-features` green, 190 suites
- `cargo clippy --all-targets --all-features -- -D warnings` clean
- `cargo fmt --all -- --check` clean

Out of scope, tracked separately: removing `Board.card_counter` itself (KAN-1216, this unblocks it) and the `CreateSubcardCommand` WIP-limit bypass.

### DIFF
--- a/crates/kanban-backend-memory/src/in_memory_store/archived_cards.rs
+++ b/crates/kanban-backend-memory/src/in_memory_store/archived_cards.rs
@@ -86,9 +86,9 @@ mod tests {
     #[test]
     fn test_insert_and_get_archived_card() {
         let store = InMemoryStore::new();
-        let mut board = make_board("B");
+        let board = make_board("B");
         let col = make_column(board.id, "C", 0);
-        let card = make_card(&mut board, col.id, "Card", 0);
+        let card = make_card(&board, col.id, "Card", 0);
         let card_id = card.id;
         let ac = ArchivedCard::new(card_id, uuid::Uuid::nil());
         store.insert_archived_card(ac).unwrap();
@@ -100,10 +100,10 @@ mod tests {
     #[test]
     fn test_list_archived_cards() {
         let store = InMemoryStore::new();
-        let mut board = make_board("B");
+        let board = make_board("B");
         let col = make_column(board.id, "C", 0);
-        let card1 = make_card(&mut board, col.id, "C1", 0);
-        let card2 = make_card(&mut board, col.id, "C2", 1);
+        let card1 = make_card(&board, col.id, "C1", 0);
+        let card2 = make_card(&board, col.id, "C2", 1);
         store
             .insert_archived_card(ArchivedCard::new(card1.id, uuid::Uuid::nil()))
             .unwrap();
@@ -117,10 +117,10 @@ mod tests {
     #[test]
     fn test_clear_sprint_from_archived_cards() {
         let store = InMemoryStore::new();
-        let mut board = make_board("B");
+        let board = make_board("B");
         let col = make_column(board.id, "C", 0);
         let sprint_id = Uuid::new_v4();
-        let mut card = make_card(&mut board, col.id, "Card", 0);
+        let mut card = make_card(&board, col.id, "Card", 0);
         card.sprint_id = Some(sprint_id);
         let card_id = card.id;
         let before = card.updated_at;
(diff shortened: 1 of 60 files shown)


fulsomenko:
Review found one real defect, now fixed.

## The three sort tiebreaker tests had become vacuous

`OrderedSorter`'s tiebreaker tests derived their expected order from the `card_number` fields of the very cards they sorted:

```rust
let card_numbers = (card1.card_number, card2.card_number, card3.card_number);
let expected = (card_numbers.0, card_numbers.1, card_numbers.2);
assert_eq!(order(&shuffled_a), expected);
```

Once `Card::new` stopped numbering cards those all read 0, so every assertion became `(0,0,0) == (0,0,0)` and held for any ordering the sorter produced. They never went red, so nothing prompted a fix.

Proven rather than argued: replacing the tiebreak comparator with `Ordering::Equal` — deleting the tiebreaker outright — left all three passing. After the fix they fail, and pass again once it is restored.

The one that mattered is `test_ordered_sorter_is_deterministic_when_primary_keys_tie`, the only guard on the invariant its own doc comment describes: tied cards jumping between renders when a backend yields them in hash order. That guard was gone.

Numbers are now set explicitly and `expected` is a literal, so each test states the ordering it requires instead of reading it back from its own inputs.

This is precisely the failure mode the card warned about. It slipped through because the expected value was *derived* rather than written down — worth remembering as the shape to look for whenever a constructor stops supplying a value.

## Also fixed

- A stale line on `CreateSubcardCommand::capture_inverse` claiming the board's `card_counter` stays bumped. Not true in this branch, and at the exact seam the PR changes.
- The `let board_mut = board.clone();` bindings that existed only to produce a `&mut Board`.

## Checked and clean

The review enumerated every non-test hunk in all 60 files programmatically: only `card.rs` and `dependency_commands.rs` contain production changes. Subcard equivalence confirmed field by field — same id, number, prefix, title, position, board_id, column_id and parent edge.

No changed expected literal anywhere, no `assert_eq!` loosened to `assert!`, no dropped field comparisons. The 7 deleted assertions all sit inside two deleted tests, both of which asserted the removed behaviour and are unadaptable; `board.rs` still covers `get_next_card_number`/`card_counter` directly, so no coverage was lost.

Several of the 31 explicit-number sites were done thoughtfully rather than mechanically — `composite_indexes.rs` turned `for _ in 0..3` into `for n in 1..=3` to preserve distinctness, and `view_strategy.rs` set 1/2/3 to keep a deliberately out-of-order fixture meaningful.

## Deliberately not fixed here

`upsert_board` at `dependency_commands.rs:484` now writes an unmodified board. Removing it would itself be a behaviour change, which this card forbids; it belongs to KAN-1216 along with the field it exists for.

- `cargo test --workspace --all-features` green, 190 suites
- `cargo clippy --all-targets --all-features -- -D warnings` clean
- `cargo fmt --all` applied

PR #603, opened 2026-08-17
KAN-1253 fix(domain): CreateSubcardCommand must respect the column WIP limit

fulsomenko:
## Defect

`CreateSubcardCommand::execute` never checked the target column's WIP limit, so creating a subcard could silently exceed it. Every other path that puts a card in a column checks it: `CreateCard::execute`, the archived-card restore, and `MoveCard::execute`. The subcard path was the only hole.

Verified against a real `SqliteBackend`, column WIP limit 1, one card already in the column:

```
column now has 1 card, wip limit 1
ordinary create -> Err("column ... has reached its WIP limit of 1")
SUBCARD create  -> Ok("OK")
cards now in column: 2
```

## Fix

Exactly one line of production code, at the top of `CreateSubcardCommand::execute`, matching how `CreateCard::execute` opens:

```rust
context.check_wip_limit(self.column_id, 1, &[])?;
```

An existing unit test (`test_create_subcard_command` in `crates/kanban-domain/tests/dependency_commands.rs`) exercised a column id that was never inserted into the store, which now fails the WIP lookup. Updated it to create a real `Column` so the parent/subcard setup matches production shape.

## Tests

Added `crates/kanban-service/tests/subcard_wip_limit.rs` against a real `SqliteBackend`:
- `test_create_subcard_into_a_full_column_is_rejected` — asserts both the rejection and that the column's card count is unchanged (not just the error).
- `test_create_subcard_into_a_column_with_room_still_succeeds` — non-regression twin.
- `test_a_rejected_subcard_does_not_consume_a_card_number` — pins that the number allocation rolls back with the batch.

## Test plan

- [x] Tests written first, confirmed to fail for the right reason (subcard created into a full column instead of rejected)
- [x] `cargo test --workspace --all-features`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --all`

### DIFF
--- a/crates/kanban-domain/src/commands/dependency_commands.rs
+++ b/crates/kanban-domain/src/commands/dependency_commands.rs
@@ -459,6 +459,7 @@ pub struct CreateSubcardCommand {
 
 impl CreateSubcardCommand {
     pub fn execute(&self, context: &CommandContext) -> KanbanResult<()> {
+        context.check_wip_limit(self.column_id, 1, &[])?;
         context.get_card(self.parent_id)?;
         let board = context.get_board(self.board_id)?;
         // Allocates from the prefix row, like every other card. Minting from
--- a/crates/kanban-domain/tests/dependency_commands.rs
+++ b/crates/kanban-domain/tests/dependency_commands.rs
@@ -579,14 +579,16 @@ fn test_create_subcard_command() {
     use kanban_domain::Board;
 
     let tc = TestContext::new();
-    let column_id = Uuid::new_v4();
 
     let mut board = Board::new("Test Board", None::<String>);
     board.card_prefix = Some("TEST".to_string());
     let board_id = board.id;
+    let column = kanban_domain::Column::new(board.id, "Col", 0);
+    let column_id = column.id;
     let parent = kanban_domain::Card::new(board.id, column_id, "Parent", 0);
     let parent_id = parent.id;
     tc.store.upsert_board(board).unwrap();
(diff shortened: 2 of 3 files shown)


fulsomenko:
Reviewed. The fix is correct and minimal — one line, placed first, matching how `CreateCard::execute` opens. TDD was followed: the red output shows two of the three tests failing for the right reason before the fix, with the non-regression twin passing throughout, which is exactly right.

## One thing the card did not anticipate

`CreateSubcardCommand` has no `require_column` of its own. `check_wip_limit` does `get_column(...)?.ok_or(not_found)` first, so adding it makes the WIP check this command's **only** column validation — and therefore closes a latent FK hole. Subcards could previously be created into a column that does not exist.

The evidence is the collateral breakage: the pre-existing `test_create_subcard_command` passed while addressing a bare `Uuid::new_v4()` that was never inserted as a column. It could only have passed because nothing validated it.

That behaviour is strictly better, but it arrived undocumented and untested, and it is now load-bearing — making the WIP check conditional, or moving it below the writes, would silently reopen the hole. Added `test_create_subcard_into_a_nonexistent_column_is_rejected` to pin it.

## On the collateral fix

Correct as made. It replaces a fictional column with a real one and changes no assertion, so the test now exercises what it always claimed to. Bundling it with the fix commit was the right call — that commit would not be green without it.

- `cargo test --workspace --all-features` green, 191 suites
- `cargo clippy --all-targets --all-features -- -D warnings` clean
- `cargo fmt --all` applied

PR #656, opened 2026-08-22
KAN-1302 feat(service): opt-in entity cache on KanbanContext

fulsomenko:
## Summary

- Adds a private, opt-in `Option<EntityCache>` field to `KanbanContext` plus four methods: `with_entity_cache`, `has_cache`, `invalidate`, and `resolve`.
- `resolve` is total: it returns `kanban_domain::Resolved` and never errors on a context that has not opted into caching, returning `Resolved::default()` instead.
- Wires command-triggered invalidation at the single funnel `KanbanContext::record_invalidation`, which covers `execute_with_extra`, `undo`, and `redo` in one place. No consumer crate changes.
- Extends the existing `RecordingStore` test double (`crates/kanban-service/src/read_recorder.rs`) with `CommandStore` and `KanbanBackend` implementations so it can back a `KanbanContext` in tests, rather than introducing a second test double.

## Decision: eager invalidation

Invalidation is applied eagerly, inside `record_invalidation`, rather than lazily derived from `last_invalidation` at the next `resolve` call. Reasons:

- `record_invalidation` is provably the single funnel for all three command paths (`execute_with_extra`, `undo`, `redo`), so one wire covers all of them.
- Eager invalidation keeps `last_invalidation` a pure audit field with no "consumed" flag to track.
- `EntityCache::invalidate` is a handful of cheap map removals, so there is no performance reason to defer it.

## Test plan

- [x] Seven new unit tests in `crates/kanban-service/src/context/entity_cache/tests.rs` covering: no-cache default, opt-in, cache-less resolve reads nothing, resolve-after-invalidate-all re-reads, warm-cache resolve reads nothing, entity-scoped invalidation re-reads only the invalidated ids, and command execution invalidating the cache.
- [x] `cargo test --all-features --workspace` green.
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo check --package kanban-cli --no-default-features --features json,sqlite` green.
- [x] `check-crate-list-sync` and `check-factory-compile-lock` guard scripts clean.
- [x] Diff scope confirmed against `origin/develop`: only the seven files touched by this change, none under `kanban-cli`, `kanban-mcp`, or `kanban-server`.

### DIFF
--- a/.changeset/kan-1302-context-cache-wiring.md
+++ b/.changeset/kan-1302-context-cache-wiring.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+service: `KanbanContext` gains an opt-in per-entity read cache. Surfaces call `with_entity_cache()` to enable it and `resolve(plan)` to fetch only what a plan still needs; committed command batches, undos and redos now drop the entities they touched from that cache. Contexts that do not opt in behave exactly as before.
--- a/crates/kanban-service/src/context/core.rs
+++ b/crates/kanban-service/src/context/core.rs
@@ -21,6 +21,7 @@ impl KanbanContext {
             session_id: Uuid::new_v4(),
             app_type: AppType::Unknown,
             last_invalidation: None,
+            cache: None,
         }
     }
 
--- a/crates/kanban-service/src/context/entity_cache.rs
+++ b/crates/kanban-service/src/context/entity_cache.rs
@@ -1,0 +1,2 @@
+#[cfg(test)]
+mod tests;
--- a/crates/kanban-service/src/context/entity_cache.rs
+++ b/crates/kanban-service/src/context/entity_cache.rs
@@ -1,2 +1,43 @@
+use super::KanbanContext;
+use crate::cache::EntityCache;
+use kanban_domain::{FetchPlan, Invalidation, KanbanResult, Resolved};
+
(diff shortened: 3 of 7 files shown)


fulsomenko:
Closing without merging, on an architectural decision rather than any defect in this PR.

The work here is correct and the review was clean. But KAN-1351 retires the EntityCache topology: the Model in kanban-view becomes the single world state, populated directly through `&dyn DataStore`, and `EntityCache`, `Resolved` and `Model::apply_resolved` all go away.

This PR's entire surface is that topology: `Option<EntityCache>` on KanbanContext, `with_entity_cache()`, `has_cache()`, `KanbanContext::resolve` and `KanbanContext::invalidate`. Merging it would land code scheduled for deletion, and it would also ship a live bug: `replace_backend` does not clear the cache, so a backend swap would keep serving the previous store's entities.

Nothing is lost by closing. `EntityCache` itself is already on develop as a standalone unwired module at `crates/kanban-service/src/cache/mod.rs`, so the multi-round resolve loop and the retry scoping stay available as a reference implementation for KAN-1351 to port from. Only the wiring goes, and the wiring is the part being replaced.

The branch is kept. See KAN-1351 for the target design and the reasoning.

PR #658, opened 2026-08-22
KAN-1329 feat(backend-http): implement core HTTP DataStore reads with a distinct transport error

fulsomenko:
## Summary

Turns `HttpBackend`'s all-`unsupported` `DataStore` impl into a working read path against the live kanban-server REST surface.

- Adds `KanbanError::Transport(String)` (kanban-domain) and `ErrorCode::UpstreamUnavailable` / 502 (kanban-api), classified in the existing exhaustive `KanbanError -> ApiError` match, plus `From<ApiError> for KanbanError` so a wire error can be turned back into a domain error without guessing at its structure.
- Adds a `get_json` / `get_json_list` helper pair on `HttpBackend`'s dedicated runtime. `get_json` maps a 404 to `Ok(None)`, any other non-2xx to `Err` carrying the server's error code, and a send failure to `KanbanError::Transport`. `get_json_list` follows `Page<T>.total_pages` to completion at the server's default page size.
- Implements nine core `DataStore` reads against the live server: `get_board`, `list_boards`, `get_column`, `list_columns_by_board`, `get_card`, `list_cards_by_column`, `list_cards_by_sprint`, `get_sprint`, `list_sprints_by_board`. The two-request card lists resolve their parent (column or sprint) first and return an empty list for an unknown parent, matching the in-memory contract.
- Rules explicitly on the five inherited `DataStore` defaults this card owns, so none of them can silently go live as a full-workspace scan now that their delegates answer real data: `get_card_by_board_and_number`, `list_cards_by_number` and `list_cards_by_prefix_and_number` decline under their own name; `get_card_by_sprint_and_number` and `list_cards_by_columns` keep the same shape as the inherited default but are written out explicitly with cost-stating doc comments.
- `upsert_board` / `upsert_column` / `upsert_card` / `upsert_sprint` stay `unsupported`: every available `PUT` route drops fields `DataStore::upsert_*` must write verbatim, and the path is unreachable today regardless (`with_transaction` already declines and there is no `RemoteWrites` impl).
- Removes the crate's blanket `#![allow(dead_code)]` now that `block_on`, `base_url` and `client` have production callers.

Board, sprint and card conversions are lossy in documented, tested ways: a remotely-fetched board cannot allocate sprint numbers or resolve sprint names (`BoardResponse` omits `sprint_names`/`sprint_name_used_count`/`next_sprint_number`), and a remotely-fetched sprint always renders unnamed (`SprintResponse` exposes a resolved `name` instead of `name_index`).

## Counts (re-derived)

`DataStore` trait methods, excluding the crate's own `#[cfg(test)]` module:

```
awk '/^pub trait DataStore/,/^}/' crates/kanban-domain/src/data_store.rs | grep -oP '^\s{4}fn \K\w+' | sort -u | wc -l
# 52
```

`impl DataStore for HttpBackend`, same exclusion (this crate has no `#[cfg(test)]` module in `data_store.rs`):

```
awk '/^impl DataStore for HttpBackend/,/^}/' crates/kanban-backend-http/src/data_store.rs | grep -oP '^\s{4}fn \K\w+' | sort -u | wc -l
# 47
```

`comm -23` of the two gives exactly the 5 names still inherited.

## D3 table

| # | Method | Status | Owner |
|---|---|---|---|
| 1 | `get_card_by_board_and_number` | explicit decline, own name | this card |
| 2 | `get_card_by_sprint_and_number` | explicit override, same shape as default | this card |
| 3 | `list_cards_by_number` | explicit decline, own name | this card |
| 4 | `list_cards_by_prefix_and_number` | explicit decline, own name | this card |
| 5 | `list_cards_by_columns` | explicit override, same shape as default | this card |
| 6 | `list_cards_by_column_filtered` | still inherited | split-1241-b |
| 7 | `count_cards_in_column_filtered` | still inherited | split-1241-b |
| 8 | `list_archived_cards_by_board` | still inherited | split-1241-b |
| 9 | `clear_sprint_from_archived_cards` | still inherited | split-1241-b |
| 10 | `modify_graph` | still inherited | split-1241-b |

## Reachability finding

`impl KanbanBackend for HttpBackend` declines `with_transaction` (`crates/kanban-backend-http/src/lib.rs`), and no `impl RemoteWrites for HttpBackend` exists, so the `self.backend.remote_writes().is_some()` guard at `crates/kanban-service/src/context/undo.rs` does not fire and `execute_with_extra` returns before reaching any `upsert_*`. The four `upsert_*` declines added in this PR are therefore unreachable from production today, which is consistent with the routes available (every one would otherwise be a lossy write) and is reported here rather than worked around.

## Test plan

- Every test in the handoff's red_tests list was observed failing at the Red commit (kanban-domain and kanban-api fail to compile; the backend-http suite exercises still-`unsupported` methods).
- Green commit implements the read path; full suite green locally: `cargo test --all-features --workspace`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`.
- Mutation-checked the pagination loop, the decline-name assertion, the Transport error classification and the get_json 404-vs-error branch by deliberately breaking each, confirming the relevant test failed, then restoring.

### DIFF
--- a/.changeset/kan-1329-http-datastore-core.md
+++ b/.changeset/kan-1329-http-datastore-core.md
@@ -1,0 +1,5 @@
+---
+bump: minor
+---
+
+backend-http: read boards, columns, cards and sprints from a remote kanban-server instead of declining every call, and report a transport failure (the server never answered) distinctly from an internal fault or a proven-absent record.
--- a/crates/kanban-api/src/v1/error.rs
+++ b/crates/kanban-api/src/v1/error.rs
@@ -213,4 +213,14 @@ mod tests {
         assert_eq!(ErrorCode::SerializationError.http_status(), 500);
         assert_eq!(ErrorCode::DatabaseError.http_status(), 500);
     }
+
+    #[test]
+    fn test_upstream_unavailable_maps_to_502() {
+        assert_eq!(ErrorCode::UpstreamUnavailable.http_status(), 502);
+        assert_eq!(ErrorCode::UpstreamUnavailable.to_string(), "UPSTREAM_UNAVAILABLE");
+        assert_ne!(
+            ErrorCode::UpstreamUnavailable.http_status(),
+            ErrorCode::InternalError.http_status()
+        );
+    }
 }
--- a/crates/kanban-api/src/v1/error.rs
+++ b/crates/kanban-api/src/v1/error.rs
@@ -28,6 +28,7 @@ pub enum ErrorCode {
     SerializationError,
     DatabaseError,
     InternalError,
+    UpstreamUnavailable,
 }
 
 impl std::fmt::Display for ErrorCode {
@@ -52,6 +53,7 @@ impl std::fmt::Display for ErrorCode {
             Self::SerializationError => "SERIALIZATION_ERROR",
             Self::DatabaseError => "DATABASE_ERROR",
             Self::InternalError => "INTERNAL_ERROR",
+            Self::UpstreamUnavailable => "UPSTREAM_UNAVAILABLE",
         };
         f.write_str(s)
     }
@@ -80,6 +82,7 @@ impl ErrorCode {
             | Self::SerializationError
             | Self::DatabaseError
             | Self::InternalError => 500,
+            Self::UpstreamUnavailable => 502,
         }
     }
 }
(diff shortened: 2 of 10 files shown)


fulsomenko:
Addressed both findings from the review.

1. `list_columns_by_board` and `list_sprints_by_board` now resolve the parent board first and return `Ok(vec![])` when it is absent, matching `list_cards_by_column` and `list_cards_by_sprint` on this backend and every other `DataStore` backend. I checked every `get_json_list` call site in `data_store.rs` (there are exactly five: `list_boards`, which has no parent; the two card-list methods, already guarded; and these two, which were not). Two new tests, `test_list_columns_by_board_returns_empty_for_an_unknown_board` and `test_list_sprints_by_board_returns_empty_for_an_unknown_board`, were confirmed failing against the prior implementation before the fix, and I mutation-tested the guard by reverting it locally and re-confirming both tests fail, then restored it.

2. `block_on`'s doc comment now states the panic contract the two runtime-context tests pin: it must not be called from a thread already inside a Tokio runtime, where it panics with "Cannot start a runtime from within a runtime", and an async caller reaches it through `tokio::task::spawn_blocking`.

All gates green locally: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-features --workspace`, `cargo check --package kanban-cli --no-default-features --features json,sqlite`, `check-crate-list-sync`, `check-factory-compile-lock`.

PR #679, opened 2026-08-27
fix(tui,domain): release blockers from the v0.9.0 quality gate

fulsomenko:
Fixes the two confirmed defects found by the pre-ship quality gate on release PR #557.

**TUI regression vs 0.8.1 (archive highlight)**: archiving a project from the Projects panel snapped the highlight to the first project instead of the vacated position. `delete_board` pinned the index against the pre-archive board list, so the next frame's identity-preserving resync moved the selection to the top. Fixed by resyncing via `prepare_frame` before pinning, the ordering `restore_archived_board` and `permanently_delete_archived_board` already use. The new regression test runs the frame resync the existing test skipped (it was red before the fix).

**SQLite data bug (non-ASCII prefixes)**: `Prefix::normalize` folded case with Unicode rules while SQLite compares `cards.prefix_ref` to `prefixes.name` with `COLLATE NOCASE` (ASCII-only), so a board with a non-ASCII card prefix (e.g. `ÖST`) failed the foreign key and could not store a card at all on SQLite, while JSON and in-memory accepted it. Fixed by folding ASCII-only, which matches what the FK and `idx_cards_prefix_nocase_number` can actually enforce. Pinned by a new cross-backend contract test (red on SQLite, green on JSON/in-memory before the fix; green on all three after) plus a unit pin. Safe pre-release: prefix rows are new in v0.9.0, so no shipped data carries the Unicode normalisation.

Verification: fmt clean, clippy `-D warnings` clean, contract suite green on all three backends (280 tests), domain/tui/json suites green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/fix-ascii-prefix-normalization.md
+++ b/.changeset/fix-ascii-prefix-normalization.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+domain: prefix normalisation now folds ASCII case only, matching the `COLLATE NOCASE` comparison SQLite uses for the `cards.prefix_ref` foreign key. Unicode folding desynced the stored row name from what SQLite could match, so a board with a non-ASCII card prefix (for example `ÖST`) could not store a card at all on the SQLite backend while JSON accepted it. Non-ASCII prefixes now round-trip on every backend, pinned by a new cross-backend contract test.
--- a/.changeset/fix-tui-archive-highlight.md
+++ b/.changeset/fix-tui-archive-highlight.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+tui: archiving a project from the Projects panel keeps the highlight at the vacated position again instead of snapping to the first project. The index was pinned against the pre-archive board list, so the next frame's identity-preserving resync moved the selection to the top; the handler now resyncs the list before pinning, matching the restore and permanent-delete handlers.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,8 @@ edition = "2021"
 rust-version = "1.74"
 authors = ["Max Emil Yoon Blomstervall"]
 license = "Apache-2.0"
-repository = "https://github.com/fulsomenko/kanban"
-homepage = "https://github.com/fulsomenko/kanban"
+repository = "https://github.com/kanban-rs/kanban"
+homepage = "https://github.com/kanban-rs/kanban"
 description = "A terminal-based project management solution"
 
 [workspace.dependencies]
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Kanban
 
-[![CI](https://github.com/fulsomenko/kanban/actions/workflows/ci.yml/badge.svg)](https://github.com/fulsomenko/kanban/actions/workflows/ci.yml)
+[![CI](https://github.com/kanban-rs/kanban/actions/workflows/ci.yml/badge.svg)](https://github.com/kanban-rs/kanban/actions/workflows/ci.yml)
 [![Crates.io](https://img.shields.io/crates/v/kanban-cli.svg)](https://crates.io/crates/kanban-cli)
 [![AUR](https://img.shields.io/aur/version/kanban?label=AUR)](https://aur.archlinux.org/packages/kanban)
 [![nixpkgs stable](https://repology.org/badge/version-for-repo/nix_stable_26_05/kanban.svg?header=nixpkgs%20stable)](https://search.nixos.org/packages?show=kanban&channel=26.05)
@@ -95,7 +95,7 @@ cargo install kanban-cli
 
 ### From source
 ```bash
-git clone https://github.com/fulsomenko/kanban
+git clone https://github.com/kanban-rs/kanban
 cd kanban
 cargo install --path crates/kanban-cli
 ```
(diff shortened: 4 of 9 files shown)


fulsomenko:
## Review findings (self-review, adversarial pass)

Verdict: **approve**. Both fixes are minimal, red-test-proven, and pre-release-safe (prefix rows first ship in v0.9.0, so no released data carries the old Unicode fold). The findings below are non-blocking follow-ups. Each one lists the offender, a concrete failure example, and a proposed resolution with code.

---

### Finding 1: card-identifier search still folds with Unicode rules, diverging from the new ASCII-only prefix equality

**Severity: minor (consistency). Non-blocking because each path is internally consistent.**

This PR makes `Prefix::normalize` ASCII-only so the domain agrees with SQLite's `COLLATE NOCASE`. But the card-identifier search path still Unicode-folds both sides, so the workspace now has two definitions of prefix equality.

**Offender 1** - `crates/kanban-domain/src/search/mod.rs:148` (identifier construction):

```rust
format!("{}-{}", prefix, card.card_number).to_lowercase()
```

**Offender 2** - `crates/kanban-domain/src/search/mod.rs:221` (`parse_identifier`, query side):

```rust
let lower = identifier.to_lowercase();
```

**Failure example.** A card in namespace `ÖST` with number 1:

- TUI/search path (`CardIdentifierSearcher`): query `öst-1` folds to `öst-1`, card identifier folds to `öst-1` -> **matches**.
- Indexed path (`DataStore::list_cards_by_prefix_and_number`, and SQLite's `prefix = ? COLLATE NOCASE`): `Prefix::normalize("öst") == "öst"` vs `Prefix::normalize("ÖST") == "Öst"` -> **no match**.

The same query resolves the card in one surface and misses it in another. Nothing corrupts, but the user-visible behaviour differs by code path.

**Proposed resolution.** Converge on `Prefix::normalize` as the single definition of prefix equality:

```rust
// search/mod.rs:148
format!(
    "{}-{}",
    kanban_domain::Prefix::normalize(prefix),
    card.card_number
)

// search/mod.rs, parse_identifier: normalize only the prefix segment,
// keep the number parse as-is
let prefix = Prefix::normalize(&identifier[..dash_pos]);
```

Note `parse_identifier` currently lowercases the WHOLE identifier before splitting; folding only the prefix half after the split is also slightly more correct (the number segment never needed folding). Pin with a unit test:

```rust
#[test]
fn test_identifier_matching_agrees_with_prefix_normalize_for_non_ascii() {
    // "ÖST-1" must be found by the query "Öst-1" (ASCII fold) on EVERY
    // lookup path, and "öst-1" must miss on every path, not just some.
}
```

---

### Finding 2: sprint-prefix collision validation folds with Unicode rules, over-reporting collisions

**Severity: nit (conservative failure mode only).**

**Offender** - `crates/kanban-domain/src/commands/sprint_commands.rs:212-235`, `validate_card_prefix_unique`:

```rust
let new_prefix_lower = new_prefix.to_lowercase();
...
.map(|p| p.to_lowercase())
```

**Failure example.** Board prefix `Öst`, proposed sprint override `öst`:

- Unicode fold says both are `öst` -> rejected as "sprint card_prefix must not match the board card_prefix".
- Under this PR's ASCII fold they are DIFFERENT namespaces (`Öst` vs `öst`), each with its own counter row, so the rejection is a false positive.

The error direction is safe (it refuses something harmless rather than allowing something harmful), which is why this is a nit and not a blocker.

**Proposed resolution.** Same convergence:

```rust
let new_prefix_lower = Prefix::normalize(new_prefix);
...
.map(|p| Prefix::normalize(p))
```

After Findings 1 and 2, `grep -rn "to_lowercase" crates/kanban-domain/src | grep -i prefix` should return only `Prefix::normalize` itself. Worth adding that as the acceptance criterion on the follow-up item.

---

### Finding 3: contract coverage stops at raw `upsert_card`; the service create path and sprint allocation are unpinned for non-ASCII

**Severity: nit (the FK exercised is the same either way).**

**Offender** - `crates/kanban-service/src/test_helpers/contract/prefix.rs`, `test_non_ascii_prefix_round_trips_a_card_on_every_backend` (added in this PR). It seeds the row and card by hand:

```rust
backend.as_data_store().upsert_prefix(Prefix::new("ÖST")).unwrap();
...
let mut card = Card::new(board.id, column.id, "one", 0);
card.prefix = "ÖST".to_string();
```

That pins the storage FK but not the full production path: `KanbanContext::create_card_from_spec` -> `allocate_card_number` -> row creation on demand -> card stamped with configured casing. It also leaves `allocate_sprint_number` untested for non-ASCII.

**Proposed resolution.** Add one contract case driving the real service path, mirroring the existing `test_creating_a_card_leaves_its_namespace_backed` shape:

```rust
pub async fn test_creating_a_card_in_a_non_ascii_namespace_succeeds_on_every_backend(
    factory: &BackendFactory,
) {
    // Board::new("B", Some("ÖST")), then ctx.create_card_from_spec(...):
    // assert card.prefix == "ÖST", card.card_number == 1, and exactly one
    // row named Prefix::normalize("ÖST") == "Öst" exists after reload.
}
```

and a sprint sibling asserting `allocate_sprint_number` mints from the same `Öst` row rather than forking a second one.

---

### Suggested disposition

Findings 1 and 2 are one coherent follow-up ("converge every prefix fold on `Prefix::normalize`") and Finding 3 is a small test enrichment. Proposal: track them as checklist items on the release-review card on the project board rather than blocking this PR, since all three have safe failure modes and the release-blocking defects this PR fixes are fully pinned.

PR #689, opened 2026-08-29
refactor(domain,view,tui): move the isolated Model into kanban-domain

fulsomenko:
## What

Moves the `Model` out of `kanban-view` into `kanban-domain`, and in the same commit strips its nine presentation fields into a new `kanban_view::controller::Controller`.

| what | before | after |
|---|---|---|
| `Model` | `kanban-view/src/model/`, 20 fields | `kanban-domain/src/model/`, **11** fields, re-exported as `kanban_domain::Model` |
| the four `displayed_*` partitions, `archived_board_at`, the four board-sort scalars, and the nine methods over them | `impl Model` | `kanban_view::controller::Controller` (new) |
| `Model::with_load_states(ModelLoadStates)` | did not exist | `kanban-domain`, gated `#[cfg(any(test, feature = "test-helpers"))]` |

## Why

The Model sat in `kanban-view`, above `kanban-service` in the dependency graph, so the service layer could not name it without taking a dependency on the presentation layer. `kanban-domain` is the lowest crate every consumer already reaches, and every type the Model is built from (`LoadState`, `Snapshot`, `Board`, `Card`, `DependencyGraph`) already lives there.

The move is only legitimate once the Model stops being presentation-aware, which is why the strip lands in the same commit. Moving a Model that names `kanban_core::AppConfig`, `BoardListFilter`, `filter_and_sort_boards`, `BoardSortField`, `SortOrder` and `chrono::DateTime<Utc>` into the domain crate would just relocate the problem.

`Model` and `Controller` are peers held side by side on `App`; the read direction is Controller to Model and never the reverse. `Controller::sync(&Model)` is the single rebuild entry point, and every snapshot load in `kanban-tui` now routes through the new `App::load_snapshot` so the partitions can never lag the model. The rebuild is not per frame.

`Model::load_from_snapshot` keeps its exact name and signature. `apply_resolved` and `mark_failed` move with the Model and lose their partition rebuilds only; neither has a caller outside `apply.rs`, and both now document that a future caller must follow with `Controller::sync`.

## Premise checks, re-derived at pickup

Baseline is `784d018f` (`origin/develop`), not the `baf76b30` the card was written against. Every figure was re-run.

```
$ git grep -n 'kanban_view::model::Model\|crate::model::Model' -- crates | wc -l
12                                 # card said 12, held (11 in kanban-tui + panel_titles.rs)

$ git grep -n '\.load_from_snapshot(' -- crates | grep -v 'src/model/' | wc -l
35                                 # card said 35, held
$ git grep -l '\.load_from_snapshot(' -- crates | grep -v 'src/model/' | wc -l
16                                 # card said 16, held

$ git grep -n '\.load_from_snapshot(' -- crates/kanban-tui/src | wc -l
15                                 # card said 15 across 9 files, held

$ for f in mod boards cards collections graph board_sort apply; do \
    grep -c '    fn test_' crates/kanban-view/src/model/$f.rs; done
6 13 11 5 3 9 13                   # card said 6/13/11/5/3/9/13, all held

$ git grep -n 'pub mod model;' -- crates/kanban-view/src/lib.rs
crates/kanban-view/src/lib.rs:17   # card said :17 not :19, held

$ git grep -n 'apply_resolved\|mark_failed' -- crates | grep -v 'model/apply.rs'
(empty)                            # H3 verified: no caller outside apply.rs
```

**One card claim did not hold.** The card states `crates/kanban-view/src/panel_titles.rs` "calls exactly one Model method, `model.sprints()` (`:48`)". It does not:

```
$ grep -n 'Model::default()\|load_from_snapshot\|model.sprints()' crates/kanban-view/src/panel_titles.rs
110:    fn board_with_sprints(...) -> (Board, Model) {
122:        let mut model = Model::default();
123:        model.load_from_snapshot(Snapshot::from_data(
137,147,160,205,215,224,233,242,254,265,277:  let model = Model::default();
174,188,189:  model.sprints()
```

`Model::default()` x12, `load_from_snapshot` x1, `sprints()` x3. The *disposition* is unaffected because every one of those methods stays on the Model, so the file still needed only its import line changed. The figure was wrong; the conclusion was right.

## Sweep, ticked off

- 12 `Model` import lines retargeted to `use kanban_domain::Model;` (no `kanban-view` re-export, per H4).
- 27 receiver-qualified moved-accessor sites plus 9 multi-line `.model\n.method()` sites moved from `app.model.X` / `self.model.X` to `.controller.X`, across `app/card_selection.rs`, `app/query.rs`, `app/view_management.rs`, `components/selection_dialog.rs`, `handlers/{board,card,column,popup,sprint}_handlers.rs`, `ui/board_detail.rs`, `ui/sprint_detail.rs`, `ui/dialogs/sprints.rs` and six test files.
- 15 in-crate + 9 integration-test `load_from_snapshot` sites routed through `App::load_snapshot`. The remaining 10 in `tests/model_tests.rs` are on a bare `Model` and correctly keep the `Model::load_from_snapshot` form (that file greps clean for `displayed_`/`live_cards`/`board_sort`, as the card predicted).
- `App::load_snapshot` is now the sole `.load_from_snapshot(` call site in `crates/kanban-tui/src`.

## Test dispositions applied

- **12 moved** to the Controller with their assertions unchanged: the nine in `model/board_sort.rs`, `test_board_sort_field_config_string_is_canonical` from `boards.rs`, and the two `test_displayed_*_partition_cached_on_load` rewritten as `load_from_snapshot` then `sync`.
- **1 deleted**: `test_apply_resolved_rebuilds_the_displayed_partitions`. Named subsumer: **`test_sync_rebuilds_the_partitions_after_the_model_changes`**, which is strictly stronger because it covers any Model change rather than one applier.
- **1 narrowed**: `test_apply_resolved_leaves_indexes_untouched_for_an_untouched_tier` keeps both index assertions and drops the four `displayed_*` before/after clones.
- **9 new** on the Controller and in `kanban-domain`'s `test_helpers`, plus 1 out-of-crate integration test.

`apply.rs` goes from 13 tests to 12. **Correction for KAN-1422:** its "eleven `apply_resolved` tests die with the file" is now **ten** (12 tests total, of which 10 exercise `apply_resolved`; the two `mark_failed` tests survive).

## Verification

```
Red step   kanban-domain  607 pre-existing pass, 8 new fail
           kanban-view    200 pre-existing pass, 5 new fail
           kanban-tui     model_load_state_ctor_tests fails at COMPILE, not runtime

Green      cargo test -p kanban-domain                     662 + 15 binaries, all ok
           cargo test -p kanban-domain --features test-helpers   16 binaries, all ok
           cargo test -p kanban-view                       155 ok  (140 non-model + 15 controller)
           cargo test -p kanban-tui                        65 test binaries, all ok, no flag
           cargo check -p kanban-cli -p kanban-mcp -p kanban-server   clean
           cargo clippy -p kanban-domain -p kanban-view -p kanban-tui \
             --all-targets --features kanban-domain/test-helpers -- -D warnings   clean
           cargo fmt --all --check                         clean
```

Acceptance greps:

```
$ git grep -n 'kanban_core\|AppConfig\|BoardSortField\|SortOrder\|filter_and_sort_boards\|chrono' \
    -- crates/kanban-domain/src/model/
(empty)                            # D9 isolation, mechanically checked

$ git grep -n 'displayed_\|board_sort' -- crates/kanban-domain/src/model/
(empty)                            # the presentation surface left the domain crate

$ git grep -n '\.load_from_snapshot(' -- crates/kanban-tui/src
crates/kanban-tui/src/app/view_management.rs:79    # exactly one, inside App::load_snapshot

$ git grep -n with_load_states -- crates | grep -v tests/
(definition, its doc comments and the manifest comment only; no production caller)

$ git diff --stat origin/develop -- '*Cargo.toml'
 crates/kanban-domain/Cargo.toml | 5 +++++
 crates/kanban-tui/Cargo.toml    | 1 +
```

`crates/kanban-view/Cargo.toml` is untouched, so `dependency_lock_tests::test_manifest_dependencies_stay_locked_to_the_renderer_free_set` passes unweakened, as does the `kanban-view` suite it guards.

The `test-helpers` gate was proved in both directions. Temporarily appending `pub fn _probe(_: kanban_domain::ModelLoadStates) {}` to `kanban-view/src/lib.rs` (which depends on `kanban-domain` without the feature) gives:

```
error[E0425]: cannot find type `ModelLoadStates` in crate `kanban_domain`
```

so the helper is genuinely absent from the default build, while `cargo test -p kanban-domain` with no flags still runs all eight of its tests through the `test` half of `any(test, feature = ...)`.

The two `include_str!` source guards in `model/mod.rs` were re-read after the `git mv` (H8): they read `boards.rs`, `cards.rs` and `graph.rs` relative to their own file, and all three moved into the same directory, so they still resolve to the intended files rather than passing vacuously.

## Changeset

`.changeset/kan-1359-move-isolated-model-kanban-domain.md`, `bump: minor`. The card specified `patch`; that is wrong. `.changeset/README.md` says that pre-1.0, anything that breaks a downstream crate takes `minor`, explicitly including "removing or renaming a public item". This PR removes the public module `kanban_view::model` and the public type `kanban_view::model::Model`, and moves nine public methods to a different type in a different crate. Every crate in `crates/` publishes to crates.io.

## Known deviations from the card

- The card's Red step says "the workspace builds; every new test fails". `crates/kanban-tui/tests/model_load_state_ctor_tests.rs` cannot satisfy that: it assigns a `kanban_domain::Model` to `app.model`, which is still a `kanban_view::model::Model` until the Green step lands. It is red at compile instead. Recorded rather than worked around, because the alternative would be to weaken the test.
- The Refactor step asks that every file under `model/` and `controller/` be under 300 lines. `crates/kanban-domain/src/model/apply.rs` is 434 (it was 501 before this PR; the rest is `#[cfg(test)]`). Inherited, not introduced, and KAN-1422 deletes the file.


fulsomenko:
Merged as `60cac97d` on `develop`. GitHub returned a 504 during the squash merge and created the merge commit without closing this PR, so the record is stale rather than the work being incomplete.

Verified all three commits landed:
- `a111d286` the Red tests: `ModelLoadStates` present in `crates/kanban-domain/src/model/test_helpers.rs`
- `e279ee38` the move: 7 files under `crates/kanban-domain/src/model/`
- `694e0c1d` the review fix: `test_sync_rebuilds_the_partitions_after_apply_resolved` present in `crates/kanban-view/src/controller/mod.rs`

A content diff of this branch against `develop` is empty except for five lines `develop` gained afterwards from KAN-1418.

Closing so it cannot be re-merged over newer state.

PR #714, opened 2026-08-31
KAN-1483 feat(mcp): card_batch tools resolve names from the call-scoped Model

fulsomenko:
## Summary

Retargets the eight `mcp_resolve_*` / `mcp_require_same_board` shim call sites in `crates/kanban-mcp/src/tools/card_batch.rs` at the call-scoped `Model` (built via `ToolScope`) so batch card resolution and column/sprint-in-board resolution read a planned tier of the Model instead of hitting the backend through the `McpResolve` shim. A `NotLoaded`/`Failed` tier now produces an error that names the collection (`"card list"`, `"columns of the board"`, `"sprints of the board"`) instead of a raw backend error or a "not found" answer.

## Changes

- `tool_archive_cards`, `tool_move_cards`, `tool_assign_cards_to_sprint`, `tool_assign_card_to_sprint` build a `ToolScope`, call `ctx.model_for(&scope)`, and resolve names through `resolve_cards`/`resolve_column_in_board`/`resolve_sprint_in_board` in `helpers/model_read.rs`. The two-round tools (`move_cards`, `assign_cards_to_sprint`, `assign_card_to_sprint`) sync a second round scoped to the resolved board before resolving the column/sprint.
- `tool_unassign_card_from_sprint`, and the card lookup inside `tool_assign_card_to_sprint`, keep using `KanbanOperations::resolve_card_id` directly (dropping only the `McpResolve` shim wrapper). Singular card-identifier resolution routes to an indexed backend lookup (`list_cards_by_prefix_and_number` / `list_cards_by_number`), and moving it to the Model's `card_list` tier would trade that indexed read for a full collection scan with no correctness gain, since the indexed path already surfaces a read failure as an error.
- `card_board` and `KanbanOperations::require_same_board` are left on their existing backend paths for the same reason: neither is name resolution, and `require_same_board` cannot be served by any tier `ToolScope::next_round` can request without editing `scope.rs`, which is out of this leaf's scope.
- `crates/kanban-mcp/src/tools/card_relations.rs` is unchanged. It has zero `mcp_resolve_*` call sites already (its six `ctx.resolve_card_id` calls are direct, on the indexed path), and its graph reads (`list_parents_of` / `list_children_of`) already propagate a failed read as an error, so there is no "collapses to not found" defect to fix there.
- `helpers/model_read.rs::resolve_cards` now requires the card-list tier lazily (on first non-uuid input) instead of unconditionally up front, matching its singular sibling `resolve_card`. Without this fix, a batch call made entirely of uuids (a normal LLM call shape) would plan no `card_list` tier and then trip `NotLoaded` -> `INTERNAL_ERROR`, because `ToolScope::next_round` only requests `card_list` when at least one reference is a name.
- Two `#[allow(dead_code)]` attributes added to `McpResolve::mcp_resolve_cards` / `mcp_require_same_board` in `helpers/resolvers.rs`: after this change no call site in the crate uses either method, and without the attribute they trip `-D warnings`. The shim itself (all eight methods) is left in place for KAN-1485 to remove.

## Read-shape changes worth flagging

- The sprint tools (`tool_assign_card_to_sprint`, `tool_assign_cards_to_sprint`) now force a `board_list` read in their second round, because `resolve_sprint_in_board` reads `board_by_id_state`, which is served by the flat boards tier. Previously `resolve_sprint_id` used a per-id `get_board(board_id)`. This trades one indexed board read for a whole-collection board read; `ToolScope`'s design intentionally does not try to avoid it.
- Batch identifier resolution now reads the Model's `card_list` tier, filled from the raw `DataStore::list_all_cards()`, rather than `KanbanOperations::list_all_cards` (which excludes archived-board descendants). An identifier belonging to a card under an archived board may resolve where it previously reported not-found. The full `kanban-mcp` test suite passes with this change; no test exercises that specific edge case today. KAN-1481 shipped the identical shift for boards.

## Verification

Re-derived per-file tool/`mcp_resolve_` census against `origin/develop`, anchored at the first column-zero `#[cfg(test)]` (board.rs now carries an inline test module, so an unanchored or absent truncation undercounts):

```
board: tools=9 mcp_resolve=0
card_crud: tools=10 mcp_resolve=17
card_batch: tools=5 mcp_resolve=8
card_relations: tools=4 mcp_resolve=0
column: tools=6 mcp_resolve=6
sprint: tools=9 mcp_resolve=10
transfer: tools=4 mcp_resolve=1
```

- `git grep -c 'mcp_resolve_\|mcp_require_same_board' -- crates/kanban-mcp/src/tools/card_batch.rs crates/kanban-mcp/src/tools/card_relations.rs` returns 0 for both files, unblocking KAN-1485's shim deletion.
- `git diff --stat origin/develop -- crates/kanban-mcp/src/scope.rs crates/kanban-mcp/src/context.rs crates/kanban-mcp/src/tools/card_relations.rs crates/kanban-mcp/src/tools/transfer.rs crates/kanban-mcp/tests/` is empty.
- `nix develop --command cargo test -p kanban-mcp --all-features` is green: 81 lib unit tests, 94 integration tests (including `tool_move_cards_rejects_cross_board_batch` and `tool_assign_card_to_sprint_resolves_by_name_then_mutates`, both unmodified), 9 `mcp_server_builder` tests, 1 doctest.
- `nix develop --command cargo clippy -p kanban-mcp --all-targets --all-features -- -D warnings` and `nix develop --command cargo fmt --all -- --check` are clean.
- Nine new tests (one in `helpers/model_read.rs`, eight in `tools/card_batch.rs`) were observed failing before the implementation commit and pass after it; the four fault/zero-read scenarios each have a JSON and an SQLite twin. Mutation-checked the two-round `sync_into` call in `tool_assign_card_to_sprint`: deleting it made the fault test pass vacuously until the assertion was tightened to check the injected-fault text, then deleting the line correctly failed the test; restored.

### DIFF
--- a/.changeset/kan-1483-mcp-card-batch-read-model.md
+++ b/.changeset/kan-1483-mcp-card-batch-read-model.md
@@ -1,0 +1,18 @@
+---
+bump: patch
+---
+
+mcp: retarget card_batch.rs's name-resolution call sites (tool_archive_cards,
+tool_move_cards, tool_assign_cards_to_sprint, tool_assign_card_to_sprint) from
+the McpResolve shim to a call-scoped Model built through ToolScope, and fix
+resolve_cards to require the card-list tier lazily so an all-uuid batch no
+longer trips an unfetched-tier error. A failed or unfetched column/sprint
+collection read now surfaces an error naming "columns of the board" or
+"sprints of the board" instead of a raw backend error. Singular card
+identifier resolution (tool_unassign_card_from_sprint, and the card lookup
+inside tool_assign_card_to_sprint) is left on the indexed
+KanbanOperations::resolve_card_id path rather than moved to the Model, since
+that path is already an indexed lookup and migrating it would trade an
+indexed read for a full collection scan with no correctness gain.
+card_relations.rs is unchanged: it has zero McpResolve shim call sites and its
+graph reads already surface a failed read as an error.
--- a/crates/kanban-mcp/src/helpers/model_read.rs
+++ b/crates/kanban-mcp/src/helpers/model_read.rs
@@ -525,4 +525,19 @@ mod tests {
         assert_eq!(unloaded_err.code, rmcp::model::ErrorCode::INTERNAL_ERROR);
         assert!(unloaded_err.message.contains("card list"));
     }
+
+    #[test]
+    fn test_resolve_cards_with_only_uuid_references_does_not_require_the_card_list() {
+        let a = Uuid::new_v4().to_string();
+        let b = Uuid::new_v4().to_string();
+        let ids = resolve_cards(&Model::default(), &[a.clone(), b.clone()]).unwrap();
+        assert_eq!(
+            ids,
+            vec![Uuid::parse_str(&a).unwrap(), Uuid::parse_str(&b).unwrap()]
+        );
+
+        let mixed_err = resolve_cards(&Model::default(), &[a, "KAN-5".to_string()]).unwrap_err();
+        assert_eq!(mixed_err.code, rmcp::model::ErrorCode::INTERNAL_ERROR);
+        assert!(mixed_err.message.contains("card list"));
+    }
 }
--- a/crates/kanban-mcp/src/helpers/model_read.rs
+++ b/crates/kanban-mcp/src/helpers/model_read.rs
@@ -4,7 +4,7 @@ use crate::helpers::error_mapping::kanban_err_to_mcp;
 use kanban_domain::{
     find_boards_by_name, find_columns_by_name, find_sprints_by_query_global,
     find_sprints_by_query_on_board, parse_identifier, AmbiguousMatch, BatchResolutionCause,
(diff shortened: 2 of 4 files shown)


fulsomenko:
Addressed the review finding: the `cards` tier of `ArchiveCardsRequest` and `AssignCardsToSprintRequest` was unpinned.

- Added `assert!(err.message.contains("injected fault"))` to both archive-fault tests, mirroring the fix already applied to the move_cards and assign_card_to_sprint twins.
- Added name-referenced happy-path tests (`test_archive_cards_by_name_archives_the_card_on_{json,sqlite}`, `test_assign_cards_to_sprint_by_name_assigns_the_card_on_{json,sqlite}`), resolving the seeded card's `PREFIX-NUMBER` identifier through the Model.

Mutation-checked: deleting the `cards:` line from either `ToolScoped` impl now fails the corresponding new tests (4 for archive, 2 for assign-cards-to-sprint); reverted after confirming.

New commit: 90da8380.

PR #721, opened 2026-09-01
KAN-1490 fix(tui): decline unloaded column/sprint tiers in card_handlers.rs

fulsomenko:
## Summary

Migrates the 11 production `Model::columns()` / `Model::sprints()` reads in `crates/kanban-tui/src/handlers/card_handlers.rs` onto the state-preserving accessors, so a `NotLoaded` tier is declined with an error banner instead of collapsing to an empty collection and silently behaving as "board has zero columns/sprints". `Model::columns()` and `Model::sprints()` themselves are unchanged (their deletion is a separate leaf).

Sites, classified per the parent epic's three replacement shapes:
- shape (a), board-scoped: `create_card_target_column` flat-column lookup (`:43`), `handle_manage_children_from_list` (`:948`)
- shape (b), by-id: `create_card_target_column` (`:51`)
- shape (c), flat `*_state()`: `handle_create_card_key` (`:18`), `handle_assign_to_sprint_key` bulk and single-card paths (`:166`, `:179`, `:208`), `create_card` (`:394`), `handle_move_card` (`:537`), `move_selected_cards` (`:612`), `restore_card_without_reload` (`:824`)

## Fix for review feedback

The prior version of `create_card`'s guard was `self.get_focused_column_id().is_none() && !board_columns_state(bid).is_loaded() && !columns_state().is_loaded()`. Under `TaskListView::ColumnView` the active task list is `CardListId::Column(..)`, so `get_focused_column_id()` returns `Some` and the whole guard short-circuited to false regardless of load state. `create_card_target_column` then resolved the focused column id through `column_by_id_state`, which maps `NotLoaded` to `None`, so `create_card` took its `None` branch and fabricated a brand-new column.

Dropped the focus conjunct so the decline fires whenever neither `board_columns_state` nor `columns_state` is loaded, regardless of what is focused. Added a red test that puts the app in `ColumnView`, confirms a column is focused, invalidates the flat columns tier, and asserts `create_card` declines with a banner and creates neither a card nor a fabricated column. Removed the stray ticket reference from the test module's doc comment.

## Judgement calls (documented, not settled by the card)

- `handle_create_card_key` and `handle_assign_to_sprint_key` abort the whole handler on a declined tier rather than silently skipping only the picker priming, since a half-primed dialog is worse UX than none.
- `column_by_id_state` at `:51` sits inside a plain `Option` chain where `Missing` and `NotLoaded` collapse to the same `None`; the divergence test (`test_create_card_target_column_prefers_board_scoped_state_over_stale_flat_tier`) covers the meaningful case (scoped Loaded vs. stale flat tier) rather than a vacuous Missing-vs-NotLoaded split.

## Known gap (out of this card's scope)

`App::reload_model` clears `columns_by_board`/`sprints_by_board` on every reload and nothing in kanban-tui currently repopulates them (`KanbanContext::sync` is unwired here). After this merges, the board-scoped sites (`:43`, `:948`) will decline on every real invocation until a sibling infra card wires that fetch. Flagged for whoever sequences the rest of the batch.

## Verification

- `cargo clippy -p kanban-tui --all-targets --all-features -- -D warnings`
- `cargo test -p kanban-tui --all-features`
- Census over `card_handlers.rs` for the parent epic's replacement pattern returns 0 remaining flat `Model::columns()`/`Model::sprints()` reads
- `git diff --stat` against `origin/develop` over kanban-tui/kanban-view/kanban-domain src lists only `card_handlers.rs`

### DIFF
--- a/.changeset/kan-1490-card-handler-reads-decline.md
+++ b/.changeset/kan-1490-card-handler-reads-decline.md
@@ -1,0 +1,9 @@
+---
+bump: patch
+---
+
+tui: card_handlers.rs no longer collapses a NotLoaded columns or sprints
+tier to an empty collection when creating, moving, restoring, or assigning
+a card, or when opening the manage-children picker. An unloaded tier now
+sets an error banner and leaves the prior state untouched instead of
+silently acting as if the board had zero columns or sprints.
--- a/crates/kanban-tui/src/handlers/card_handlers.rs
+++ b/crates/kanban-tui/src/handlers/card_handlers.rs
@@ -396,8 +396,7 @@ impl App {
                 .map(|b| b.id);
 
             if let Some(bid) = board_info {
-                if self.get_focused_column_id().is_none()
-                    && !self.model.board_columns_state(bid).is_loaded()
+                if !self.model.board_columns_state(bid).is_loaded()
                     && !self.model.columns_state().is_loaded()
                 {
                     self.set_error("Columns are not loaded yet");
--- a/crates/kanban-tui/src/handlers/card_handlers.rs
+++ b/crates/kanban-tui/src/handlers/card_handlers.rs
@@ -5,7 +5,7 @@ use kanban_domain::commands::{
     SetBoardTaskSort, UpdateCard,
 };
 use kanban_domain::Model;
-use kanban_domain::{ArchivedCard, CardStatus, CardUpdate, KanbanOperations};
+use kanban_domain::{ArchivedCard, CardStatus, CardUpdate, KanbanOperations, LoadState};
 use kanban_view::card_list::CardListId;
 use ratatui::{backend::CrosstermBackend, Terminal};
 use std::io;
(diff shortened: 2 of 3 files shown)


fulsomenko:
Addressed both findings from the review:

1. The `create_card` decline guard at `card_handlers.rs:399` no longer conjuncts on `get_focused_column_id().is_none()`. It now runs unconditionally, so a `NotLoaded` columns tier declines even with a column focused (`ColumnView`). Added `test_create_card_declines_on_a_not_loaded_column_tier_with_a_column_focused`, which switches to `TaskListView::ColumnView`, confirms the active list is `CardListId::Column`, invalidates the flat columns tier, and asserts the banner is set and no column/card is fabricated. Also added `test_create_card_target_column_distinguishes_missing_from_not_loaded`, which pins that `column_by_id_state` returns distinct `LoadState::Missing` vs `LoadState::NotLoaded` for a genuinely-absent id versus a stale one, even though both currently collapse to `None` through `.loaded()` at that shape-b call site (the guard at :399 is what declines before this helper is ever reached; the helper itself stays self-only per the original scope decision).
2. Dropped the `KAN-1490` reference from the test-file module doc comment.

Mutation-checked the guard fix: reintroducing the `get_focused_column_id().is_none() &&` conjunct makes `test_create_card_declines_on_a_not_loaded_column_tier_with_a_column_focused` fail with `declining a NotLoaded columns tier must set an error banner even with a column focused`, confirming the test actually pins the fix.

PR #722, opened 2026-09-01
KAN-1491 fix(tui): column/board handlers decline on unloaded scoped tiers

fulsomenko:
## What changed

Migrates the 9 `model.columns()`/`model.sprints()` reads in
`crates/kanban-tui/src/handlers/column_handlers.rs` (7 sites) and
`crates/kanban-tui/src/handlers/board_handlers.rs` (2 sites) onto the
board-scoped, state-preserving accessors (`board_columns_state` /
`board_sprints_state`), so an unloaded tier declines with an error banner
instead of silently taking the empty-collection branch.

Affected handlers: `handle_delete_column_key`, `handle_move_column_up`,
`handle_move_column_down`, `create_column`, `delete_column`,
`handle_delete_board_key`, `handle_delete_archived_board_key`.

`board_delete_counts` (`board_handlers.rs`) is a `&self` helper that cannot
decline itself in place; its signature changes from `-> BoardDeleteCounts`
to `-> Option<BoardDeleteCounts>`, and its two callers now show an error
banner when it returns `None`. This is a third `&self`-helper exception
beyond the two the parent card names (`create_card_target_column`,
`visible_board_columns`); flagging it explicitly rather than folding it
into the "mechanical" bucket.

## Fix for the second review rejection

The prior revision added `App::sync_board_scoped_tiers`, called after
`Model::load_from_snapshot`, to repopulate `columns_by_board`/
`sprints_by_board` (which `load_from_snapshot` clears but never refills) so
the migrated handlers stay functional on a real reload. That fix was itself
wrong: it read the flat boards/columns/sprints tiers with `loaded_or_empty`,
so a `NotLoaded` flat columns or sprints tier silently became
`Loaded([])` on every board's scoped tier the moment boards were loaded.
`board_columns_state`/`board_sprints_state` then reported `Loaded` empty for
data that was never fetched, and the handlers this leaf migrated took the
zero-item branch instead of declining. The same `loaded_or_empty` calls were
also two new production `model.columns()`/`model.sprints()` reads in
`app/view_management.rs`, a file this leaf does not own, which regressed the
epic's site count from -9 to -7.

This revision fixes the seam itself: `sync_board_scoped_tiers` now
destructures `boards_state()`/`columns_state()`/`sprints_state()` against
`LoadState::Loaded` directly and returns `None` without touching the scoped
tiers unless all three flat tiers are themselves `Loaded`. Its return type
changed from `ModelChanged` to `Option<ModelChanged>`, and `load_snapshot`
merges the receipt only when it is `Some`. A new test,
`test_sync_board_scoped_tiers_leaves_not_loaded_column_tier_not_loaded`,
builds a `Model` with `boards: Loaded` and the columns/sprints tiers left at
their default `NotLoaded`, drives it through the seam directly, and asserts
`board_columns_state`/`board_sprints_state` are still `NotLoaded` rather than
`Loaded([])`. Verified by mutation: reverting the guard back to
`loaded_or_empty` makes this test fail; the fixed guard passes it.

`app/view_management.rs` is back to 2 production `model.columns()`/
`model.sprints()` reads (the pre-existing `prepare_frame` pair owned by a
sibling leaf), matching origin/develop; `column_handlers.rs` and
`board_handlers.rs` are at 0.

## Out of scope

`Model::columns()`/`Model::sprints()` and the collections module in
`kanban-domain` are untouched; this stays additive plus call-site rewiring
in `kanban-tui`, aside from the `sync_board_scoped_tiers` helper in
`App::load_snapshot`, which exists to keep the migrated handlers functional
on a real reload rather than deferring the whole leaf.

## Tests

```
cargo test -p kanban-tui --all-features
cargo clippy --all-targets --all-features -- -D warnings
cargo fmt --all -- --check
```

All green.

### DIFF
--- a/.changeset/kan-1491-column-board-handler-decline.md
+++ b/.changeset/kan-1491-column-board-handler-decline.md
@@ -5,4 +5,6 @@ bump: patch
 kanban-tui: column and board delete/move/create handlers now decline with an
 error banner when the board-scoped column or sprint tier is not loaded,
 instead of silently acting on an empty collection derived from the flat
-model tier.
+model tier. `App::load_snapshot` now repopulates the board-scoped column and
+sprint tiers from the flat snapshot it loads, so the declining handlers stay
+functional on every real reload instead of always taking the decline branch.
--- a/.changeset/kan-1491-column-board-handler-decline.md
+++ b/.changeset/kan-1491-column-board-handler-decline.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+---
+
+kanban-tui: column and board delete/move/create handlers now decline with an
+error banner when the board-scoped column or sprint tier is not loaded,
+instead of silently acting on an empty collection derived from the flat
+model tier.
--- a/crates/kanban-tui/src/app/view_management.rs
+++ b/crates/kanban-tui/src/app/view_management.rs
@@ -80,25 +80,32 @@ impl App {
     /// partitions can never lag the model.
     pub fn load_snapshot(&mut self, snapshot: Snapshot) {
         let changed = self.model.load_from_snapshot(snapshot);
-        let scope_changed = self.sync_board_scoped_tiers();
-        self.controller
-            .resync(&self.model, changed.merge(scope_changed));
+        let changed = match self.sync_board_scoped_tiers() {
+            Some(scope_changed) => changed.merge(scope_changed),
+            None => changed,
+        };
+        self.controller.resync(&self.model, changed);
     }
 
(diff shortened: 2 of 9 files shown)


fulsomenko:
Superseded. The production gap this branch fixed in view_management.rs is being fixed at the root by KAN-1495 (load_from_snapshot populates the board-scoped tiers). The handler migration will be redone on KAN-1491/tui-column-board-handlers-decline-v2 once that lands. Branch kept for reference.

PR #732, opened 2026-09-01
KAN-1429 feat(tui): render the first frame on two scoped populate passes

fulsomenko:
## Summary

- `App::load_initial_state` runs two scoped `populate` passes over the board list and the auto-selected board's subtree instead of one whole-store `ctx.snapshot()` read
- A failed board-list read still clears the save file and raises the error banner, now driven off `Model::boards_state()` being `LoadState::Failed` instead of a `Result` from `snapshot()`
- Adds `App::populate(ViewScope)`, the App-level wrapper around `TuiContext::sync`
- `check_ended_sprints` declines on a `NotLoaded` sprint tier instead of scanning it through the collapsing `Model::sprints()` accessor, and now returns `Option<Vec<Uuid>>` so a decline (`None`) is distinguishable from a scan that found nothing (`Some(vec![])`)
- Adds `startup_lazy_load_tests.rs` covering the board-list-then-subtree fetch order, per-board isolation of the by-parent tiers, the failed-read banner path, and the sprint-log migration ordering

## Fixes from the previous review round

- Adds backend-parity tests (`backend_parity::test_startup_scope_reads_the_same_rows_on_every_backend`, `...reports_the_same_variant_for_an_empty_board_on_every_backend`) driving the pass-1/pass-2 `ViewScope` through `KanbanContext::resolve` on `InMemoryStore`, `JsonDataStore` and `SqliteBackend`, all seeded from one shared `Snapshot` so timestamps cannot drift between backends. Compares boards, columns, cards (via `kanban_service::test_helpers::contract::assert_card_eq`), sprints and the dependency graph, plus the `LoadState` variant reported for an empty board's column/sprint/card tiers on every backend.
- `check_ended_sprints` now returns `Option<Vec<Uuid>>` instead of `Vec<Uuid>`, so the decline test asserts `None` rather than an empty vec that a mutated (non-declining) implementation would also produce. Mutation-checked: reverting the guard to the old `Model::sprints()` accessor now fails the decline test.
- The failed-read banner test asserts the banner's message text contains both "Failed to read data file" and the injected fault, not just that a banner exists. Adds a characterisation test against a genuinely corrupt JSON file read through `JsonDataStore`, not just an injected backend fault.
- Rewrites the migration-ordering test to seed a card whose `sprint_id` is set but whose `sprint_logs` are stale, and assert the migrated log is visible in the first fetch. The `mark_clean` assertion now lives in its own separately named test.

## Test plan

- [x] `cargo test -p kanban-tui --all-features` green (all 11 tests in `startup_lazy_load_tests.rs`, including the two backend-parity tests, plus the untouched `initial_board_selection_tests.rs` and `frame_reload_tests.rs`)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean across the workspace
- [x] `cargo fmt --all -- --check` clean
- [x] Confirmed each new/rewritten test fails for the stated reason against the pre-fix code, then reverted
- [x] Mutation-checked `check_ended_sprints`'s `Option` return: reverting the guard to `Model::sprints()` makes the decline test fail with `left: Some([]), right: None`
- [x] Mutation-checked the migration-ordering test: moving `migrate_sprint_logs()` after both `populate` calls makes it fail

## Archived views: reworked per design ruling

An earlier round made startup itself populate the archival marker tiers, via a new `App::populate_archival_markers` that fetched `list_archived_cards`/`list_archived_boards` and then resolved every marker's board/card body one id at a time, merging the results into the Model through three new domain methods. That was rejected at design level: per-id body fetches are exactly the unbounded eager-loading pattern this epic exists to remove, and the merges silently no-op whenever the flat board/card tier is still `NotLoaded`, which is precisely the state an empty archived view needs to recover from.

The mechanism is replaced:

- Startup loads nothing archived. `App::populate_archival_markers` and the domain's `merge_archived_cards`, `merge_archived_boards` and `set_archival_markers` are deleted.
- Entering an archived view populates it through the existing `reload_model` snapshot path instead. `handle_toggle_archived_cards_view` and `handle_toggle_archived_boards_view` call `self.reload_model()` on entry, before their existing `prepare_frame()`, gated on the archival tier never having been absorbed (`Model::archived_boards_absorbed()` / `Model::archived_card_markers_absorbed()`, two new predicates that distinguish "never loaded" from "genuinely empty"). A snapshot legitimately carries markers and bodies together in one read, so this is honest where the per-id merge was not; it is strictly better than `develop`, which paid the whole-store read on every launch, not just on first entry to an archived view; and it is self-healing after archive/restore mutations, which already reload. `prepare_frame` stays pure, and `ViewScope` is untouched: it keeps no archived fields.
- The headline op-log claim is re-pinned: the startup op log now carries no `snapshot`-class op and no archival read at all, and at most the existing bound of whole-store list reads carried over from `ViewScope`'s transitional flat arms and `migrate_sprint_logs`.
- New probe test `test_entering_an_archived_view_after_startup_displays_its_entity`: seeds one board, one column and one card, archives the card then the board (no live board remains), runs `load_initial_state`, then enters each archived view in turn and asserts it renders its entity. Fails against the deleted startup-population mechanism; passes against the entry-reload gate.

## Test plan (this round)

- [x] `cargo test -p kanban-tui -p kanban-domain --all-features` green, including the new probe test and every previously-approved test (two-pass startup, backend parity, failed-read banner text, migration ordering, `check_ended_sprints` decline)
- [x] Mutation-checked the entry-reload gate: removing the `reload_model()` call from `handle_toggle_archived_cards_view` makes the probe test fail; restored
- [x] `cargo clippy -p kanban-tui -p kanban-domain --all-targets --all-features -- -D warnings` clean
- [x] `cargo fmt --check` clean

### DIFF
--- a/.changeset/kan-1429-render-first-frame-partially-loaded.md
+++ b/.changeset/kan-1429-render-first-frame-partially-loaded.md
@@ -1,0 +1,5 @@
+---
+bump: minor
+---
+
+tui: `App::load_initial_state` now runs two scoped `populate` passes over the board list and the auto-selected board's subtree instead of one whole-store `snapshot` read, and `check_ended_sprints` declines on a `NotLoaded` sprint tier instead of scanning it through the collapsing `Model::sprints()` accessor.
--- a/.changeset/kan-1429-render-first-frame-partially-loaded.md
+++ b/.changeset/kan-1429-render-first-frame-partially-loaded.md
@@ -2,4 +2,4 @@
 bump: minor
 ---
 
-tui: `App::load_initial_state` now runs two scoped `populate` passes over the board list and the auto-selected board's subtree instead of one whole-store `snapshot` read, and `check_ended_sprints` declines on a `NotLoaded` sprint tier instead of scanning it through the collapsing `Model::sprints()` accessor.
+domain, tui: `App::load_initial_state` now runs two scoped `populate` passes over the board list and the auto-selected board's subtree instead of one whole-store `snapshot` read, and `check_ended_sprints` declines on a `NotLoaded` sprint tier instead of scanning it through the collapsing `Model::sprints()` accessor. Startup no longer loads archival markers at all; entering the archived-boards or archived-cards view instead triggers a `reload_model` snapshot on first entry, gated by the new `Model::archived_boards_absorbed` / `Model::archived_card_markers_absorbed` predicates so a later mutation-driven reload is not repeated.
--- a/crates/kanban-domain/src/model/boards.rs
+++ b/crates/kanban-domain/src/model/boards.rs
@@ -21,6 +21,12 @@ impl Model {
         self.archived_boards.as_deref().unwrap_or(&[])
     }
 
+    /// Distinguishes a genuinely empty archived-boards tier from one that has
+    /// never been absorbed (`archived_boards()` returns `&[]` for both).
+    pub fn archived_boards_absorbed(&self) -> bool {
+        self.archived_boards.is_some()
+    }
+
     /// Ids of the archived boards. The heads themselves live in the unified
     /// `boards_state()` collection; this set records which of them are archived (built
     /// from the markers). The live/archived partition is a presentation concern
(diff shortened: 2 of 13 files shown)


fulsomenko:
On hold after the third review rejection. The remaining findings are design-level, not implementation-level: archived-card bodies have no resolve-driven population path (they live in the flat cards tier, which pass 2 never requests when no live board is in scope), and the headline no-whole-store claim is unsatisfiable while ViewScope deliberately requests the flat tiers (the documented transitional ruling). The card tree is being corrected before any further attempt; do not merge this branch as is.

PR #747, opened 2026-09-02
KAN-1457 feat(service): add InvalidationPlan, a FetchPlan derived from an Invalidation

fulsomenko:
## Summary

- Adds `InvalidationPlan` to `kanban-service`: a `FetchPlan` built from an `Invalidation` plus a pre-invalidation `LoadedState` snapshot.
- `for_invalidation` re-requests exactly the tiers `Model::invalidate` is about to blank, restricted to tiers the model had already read (`FetchStatus::Loaded`). It returns `None` when nothing that was actually read would be affected, so a caller never pays for an empty round.
- `Invalidation::All` repairs every flat tier the snapshot shows was read, including both flat archival tiers (`archived_card_list`, `archived_board_list`), matching `Model::invalidate`'s full reset on that path.
- `Invalidation::Entities(ids)` never repairs the two flat archival tiers, because `Model::invalidate`'s `Entities` branch never blanks them (only `load_from_snapshot` recomputes those).
- The plan never emits any of the four parent-scoped tiers (`columns_by_board`, `cards_by_column`, `sprints_by_board`, `archived_cards_by_board`), since a parent key is not recoverable from the id-keyed `LoadedState` interface.
- `next_round` (the `FetchPlan` impl) re-masks every tier against a fresh `LoadedEntities` snapshot at resolve time, so anything that got loaded in the meantime drops out.
- Pure new type with 15 unit tests; zero consumer wiring in this PR (a follow-up wires it into the mutation paths) and zero change to `kanban-domain`.

## Test plan

- `cargo test -p kanban-service --all-features`: all green (194 lib tests including 15 new `invalidation_plan` tests, plus the full integration suite).
- `cargo clippy -p kanban-service --all-targets --all-features -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- Mutation check: reverted the hardcoded `archived_card_list: false` under the `Entities` arm to a `was_read`-gated version, confirmed `test_an_invalidated_archived_card_list_is_not_requested_because_invalidate_does_not_blank_it` fails, then restored the fix.

### DIFF
--- a/.changeset/kan-1457-invalidation-plan.md
+++ b/.changeset/kan-1457-invalidation-plan.md
@@ -1,0 +1,5 @@
+---
+bump: minor
+---
+
+kanban-service adds `InvalidationPlan`, a `FetchPlan` built from an `Invalidation` plus a pre-invalidation `LoadedState` snapshot. It re-requests exactly the tiers `Model::invalidate` is about to blank, restricted to tiers the model had already read, so a mutation only refetches what was actually visible before it ran.
--- a/crates/kanban-service/src/invalidation_plan.rs
+++ b/crates/kanban-service/src/invalidation_plan.rs
@@ -1,6 +1,8 @@
-use kanban_domain::Invalidation;
+use kanban_domain::{EntityIds, Invalidation};
 
-use crate::fetch_plan::{FetchPlan, FetchRound, LoadedEntities, LoadedState};
+use crate::fetch_plan::{
+    requestable, FetchPlan, FetchRound, FetchStatus, LoadedEntities, LoadedState,
+};
 
 /// A [`FetchPlan`] built from an [`Invalidation`] plus a pre-invalidation
 /// [`LoadedState`] snapshot. Construct it *before* calling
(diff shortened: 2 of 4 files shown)


fulsomenko:
Addressed both findings from the review:

1. `was_read` restored to `!matches!(status, FetchStatus::NotLoaded)` (it had narrowed to `Loaded`-only, so a `Failed` flat tier or a `Missing` per-id tier was never re-requested even though `Model::invalidate` treats both as having been read). Added two new red tests that fail against the narrowed predicate and pass against the restored one: `test_a_failed_card_list_is_still_re_requested_because_it_was_read` and `test_a_missing_card_id_is_still_re_requested_because_it_was_read`. Mutation-checked: reverting the predicate makes exactly those two tests fail, nothing else.
2. Restored the three doc comments dropped from the shipped file: the `was_read` semantics comment, the `for_invalidation` contract (pre-invalidate `loaded` snapshot requirement, `None` vs empty-plan semantics), and the `next_round` exhaustiveness contract (why both `FetchRound` literals are fully named and must never collapse to `..Default::default()`).

All 17 invalidation_plan tests plus the full `kanban-service --all-features` suite are green, `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` are clean.

PR #750, opened 2026-09-02
KAN-1442 feat(tui): narrow the migration probe to list_boards, drop write-back hydration

fulsomenko:
## Summary

- `handle_migration_complete`'s pre-swap readability probe now calls `list_boards()` instead of reading a whole `Snapshot` via `new_backend.snapshot()`.
- The sort-field/order sync that used to ride the write-back (`snapshot.apply_to_app(self)`) now reads directly from the new backend after the swap, via `self.ctx.data_store().get_board(board_id)`.
- This removes the write-back's only remaining caller in the TUI, so the now-dead `TuiSnapshot` trait and `crates/kanban-tui/src/state/snapshot.rs` are deleted entirely.
- Its serialization round-trip test relocates to `kanban-persistence::snapshot_serde`, where `snapshot_to_json_bytes`/`snapshot_from_json_bytes` already live.
- Adds `FailingBoardListBackend` to `kanban-tui`'s test helpers (a `list_boards()`-failing sibling of `FailingSnapshotBackend`) and retargets the two existing post-swap-failure tests onto it, since the probe they exercise no longer calls `snapshot()`.

## Notes

- `AlwaysFailsSnapshotBackend` was deleted outright by KAN-1415 (not relocated); this PR does not touch it.
- Changeset bump is `minor`, not `patch`: this removes a public item (`kanban_tui::state::TuiSnapshot`), which `.changeset/README.md`'s pre-1.0 rule requires `minor` for.
- `crates/kanban-tui/src/tui_context.rs` is untouched by this PR; `TuiContext::snapshot`/`apply_snapshot`/`data_store` still have other callers and are left for a later card.

## Test plan

- [x] Red: added/retargeted tests fail against the unmodified probe (see commit `test(tui): pin migration probe and write-back behavior ahead of rewrite`); characterisation tests (sort-field sync, model population, full-graph json/sqlite pins) already pass unmodified, as expected of a pin.
- [x] Green: `nix develop --command cargo test -p kanban-tui -p kanban-persistence --all-features` all green.
- [x] Mutation check: reverting the `list_boards()` abort check breaks both retargeted abort/save-worker tests; restored before commit.
- [x] `nix develop --command cargo fmt --all -- --check` clean.
- [x] `nix develop --command cargo clippy --all-targets --all-features -- -D warnings` clean.
- [x] `git grep -nE 'TuiSnapshot|apply_to_app' -- crates/kanban-tui` returns nothing.
- [x] `git grep -n 'mod snapshot' -- crates/kanban-tui/src/state/mod.rs` returns nothing.
- [x] `git grep -nE 'snapshot\(\)|apply_snapshot\(' -- crates/kanban-tui/src/handlers/settings_handlers.rs` returns nothing.


fulsomenko:
Addressed both review findings:

1. Both abort tests now capture the outgoing backend's `instance_id()` before calling `handle_migration_complete` and assert it is unchanged afterward, pinning that the swap never happened. Mutation-checked: moving the `list_boards()` probe to after `replace_backend()` makes both tests fail with a clear `left/right` instance_id mismatch; reverted after confirming.
2. Both tests now assert `banner.message.contains("simulated transient read failure")`, pinning that the abort was caused by the injected `FailingBoardListBackend::list_boards` fault rather than an unrelated earlier error path (e.g. `make_backend` failing).

`cargo test -p kanban-tui -p kanban-persistence`, `cargo fmt --all -- --check`, and `cargo clippy --all-targets --all-features -- -D warnings` all pass.

PR #757, opened 2026-09-02
KAN-1441 feat(tui): route adopt_storage_file through transfer_state_to

fulsomenko:
## Summary

- `TuiContext::adopt_storage_file` (the storage-adoption dialog's "confirm" path) now routes through the new `KanbanContext::transfer_state_to` instead of a whole-store `ctx.snapshot()` + `backend.apply_snapshot()` round trip.
- Added `TuiContext::transfer_state_to`, a thin pass-through to `KanbanContext::transfer_state_to`, placed beside `fn snapshot`.
- The post-seed readability probe changes from `backend.snapshot()` to the cheaper `backend.list_boards()`.
- The seeded backend is explicitly marked dirty with `backend.mark_dirty()` after the transfer, so the queued flush writes to disk even for a source workspace with nothing to upsert.
- `apply_snapshot` and `TuiContext::snapshot` are left in place; other call sites and a separate card own their removal.

## Test plan

New/updated tests in `crates/kanban-tui/src/app/lifecycle_tests.rs`, all `#[tokio::test(flavor = "multi_thread")]`:

- `test_adopt_storage_file_writes_the_whole_workspace_to_disk_json` / `_sqlite`: characterisation baseline, now on both backends. Both seed a non-trivial graph (3 boards with distinct card prefixes, one archived with its own column/card subtree, 4 columns each with a distinct non-default `wip_limit`, 5 cards across the boards, a sprint bound to a card, two block edges of different severities where one endpoint is an archived card, an archived-card marker, an archived-board marker, and three prefix rows with non-zero counters plus the default `sprint` prefix row). Every entity is asserted individually by id and field, never by a bare collection length. The JSON leg reopens the written file through a fresh `JsonDataStore`; the SQLite leg reopens it through a fresh `SqliteBackendFactory`-built backend, so both legs exercise a real cross-process reload rather than reading through the same in-process handle.
- `test_adopt_storage_file_writes_an_empty_workspace_to_disk`: characterisation baseline for a zero-entity workspace. Was green before this change, still green after.
- `test_adopt_storage_file_does_not_call_the_whole_store_trait_methods`: genuine Red before this change. A `HostileSourceBackend` (errors on `DataStore::snapshot`) and a `HostileTargetBackend` (errors on `DataStore::apply_snapshot`), installed as the source backend and via a custom `StoreManager` factory for the adopted target, made `adopt_storage_file` fail. Green after the fix, since neither whole-store method is called anymore.
- `test_adopt_storage_file_aborts_without_swapping_when_the_target_is_unreadable`: genuine Red before this change. An `UnreadableTargetBackend` that errors on `list_boards` did not affect the old code, which probed with `backend.snapshot()` instead. Green after: `adopt_storage_file` returns `false`, the backend is not swapped (`instance_id` unchanged), `save_file` stays unset, and the banner is `Error`-variant with a message containing `"UnreadableTargetBackend: list_boards"`, pinning the read-back probe specifically as the cause of the abort.
- `test_adopt_storage_file_still_refuses_an_existing_path`: regression pin for the untouched existing-path guard, now also pinning the banner's `Error` variant and `"already exists"` message rather than only checking a banner is present.

Mutation-checked the whole-workspace pin: temporarily made `write_full_snapshot`'s column loop rewrite every column's `name` to `"CLOBBERED"` and clear its `wip_limit`. Both `_json` and `_sqlite` legs failed on the per-column field assertions (`left: "CLOBBERED", right: "Todo"`), confirming the assertions are load-bearing rather than checking only collection lengths. Reverted before committing; `git diff -- crates/kanban-service/src/store_adapter.rs` is empty.

Separately confirmed by mutation that `backend.mark_dirty()` is not exercised by the empty-workspace test on the JSON backend: removing the call left `test_adopt_storage_file_writes_an_empty_workspace_to_disk` green, because `write_full_snapshot`'s final `set_graph` call unconditionally dirties the JSON backend regardless of whether anything else was written. `mark_dirty()` is kept as a deliberate defensive call (per the card's DECIDED #1) for backends where an empty write does not otherwise dirty; the code comment states only that fact, not a false claim about the JSON backend's current behavior.

Verification (all via `nix develop --command`, `CARGO_TARGET_DIR`/`CARGO_INCREMENTAL=0`/`CARGO_PROFILE_DEV_DEBUG=0` set):

- `cargo clippy --all-targets --all-features -p kanban-tui -- -D warnings`
- `cargo test -p kanban-tui --all-features`

All green.

### DIFF
--- a/.changeset/kan-1441-seed-adopted-storage-file-transfer-state-to.md
+++ b/.changeset/kan-1441-seed-adopted-storage-file-transfer-state-to.md
@@ -1,0 +1,5 @@
+---
+bump: minor
+---
+
+kanban-tui: route the storage-adoption dialog's `adopt_storage_file` through the new `TuiContext::transfer_state_to` instead of a whole-store `snapshot`/`apply_snapshot` round trip, replace the post-seed readability probe with `backend.list_boards()`, and explicitly `mark_dirty()` the seeded backend so the queued flush writes to disk even for an empty workspace.
--- a/crates/kanban-tui/src/app/lifecycle.rs
+++ b/crates/kanban-tui/src/app/lifecycle.rs
@@ -263,16 +263,6 @@ impl App {
             "adopt_storage_file requires a multi-threaded Tokio runtime; \
              block_in_place is unavailable on a current_thread runtime."
         );
-        // Capture in-memory state before swapping backends; replace_backend
-        // discards the old backend and the new one starts empty (or loaded
-        // from a non-existent file).
-        let snapshot = match self.ctx.snapshot() {
-            Ok(s) => s,
-            Err(e) => {
-                self.set_error(format!("Could not capture in-memory state: {}", e));
-                return false;
-            }
-        };
 
         let store_manager = self.store_manager.clone();
         let app_config = self.app_config.clone();
@@ -291,15 +281,17 @@ impl App {
 
         match backend_result {
             Ok(backend) => {
-                // Transfer the in-memory state to the new backend; this also
-                // marks it dirty so the queued flush actually writes to disk.
-                if let Err(e) = backend.as_data_store().apply_snapshot(snapshot) {
+                // transfer_state_to reads and writes the workspace together, so
+                // upserts of an empty source leave the target's dirty flag
+                // unset; mark_dirty forces the queued flush to write anyway.
+                if let Err(e) = self.ctx.transfer_state_to(backend.as_data_store()) {
                     self.set_error(format!("Could not seed \"{}\": {}", filename, e));
                     return false;
                 }
+                backend.mark_dirty();
                 // Probe the read paths so any backend failure surfaces
                 // before we commit by swapping the backend in.
-                if let Err(e) = backend.snapshot() {
+                if let Err(e) = backend.list_boards() {
                     self.set_error(format!(
                         "Could not read seeded snapshot from \"{}\": {}",
                         filename, e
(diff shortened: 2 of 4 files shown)


fulsomenko:
Addressed all four rejection findings:

1. Added a `.sqlite3` leg of the whole-workspace pin (`test_adopt_storage_file_writes_the_whole_workspace_to_disk_sqlite`), reading back through a freshly opened `SqliteBackendFactory`-built backend, sharing the same seed/assertion helpers as the `.json` leg.
2. Replaced every bare collection-length assertion in the pin test with per-entity, per-field assertions (board names; column name/position/board_id/wip_limit; card title/column_id/position/prefix/card_number/sprint_id; archived-card and archived-board markers by entity_id; both block edges by endpoint and severity; prefix rows by name and counter). Extended the seeded graph per the card's own gaps: a third board with its own subtree, archived; a second block edge with an archived target endpoint. Mutation-checked by temporarily clobbering every column's name/wip_limit in `write_full_snapshot` and confirming both legs fail before reverting.
3. Deleted the false three-line comment above `mark_dirty()` and replaced it with a one-line comment stating only what mutation-testing confirmed is true. Fixed the read-back probe's error string to `"Could not read back \"{}\": {}"`.
4. Both the abort-on-unreadable-target and refuse-existing-path tests now bind the banner, assert `BannerVariant::Error`, and assert the message contains the discriminating substring (`"UnreadableTargetBackend: list_boards"` / `"already exists"`) instead of only checking `banner.is_some()`.

`cargo test -p kanban-tui --all-features`, `cargo clippy --all-targets --all-features -p kanban-tui -- -D warnings`, and `cargo fmt --all -- --check` are all clean.

PR #761, opened 2026-09-02
KAN-1498 feat(tui): migrate cards/boards tier reads off loaded_or_empty

fulsomenko:
## Summary

Migrates the 19 remaining production `loaded_or_empty()` reads on the cards/boards tiers in kanban-tui's handlers/app onto explicit `LoadState` handling.

- Acting handlers (`create_card`, `handle_move_card`, `move_selected_cards`, `toggle_selected_cards_completion`, `start_delete_animation`, `handle_manage_children_from_list`, `CardListAction::MoveColumn`, `handle_manage_parents`, `handle_manage_children`, `toggle_completion_for_card_ids`) decline with a "not loaded yet" banner and leave state unchanged when their cards tier is not loaded.
- Per-frame paths (`handle_animation_tick`'s archive branch, `check_ended_sprints`'s board-name lookup) and read-only per-keystroke search filters (the relationship popup) skip silently, matching the established per-frame decline shape.
- `cursor_archive_anchor` and the carry-over-sprint label lookup keep their existing no-banner, caller-declines/degrades contract, just made explicit.
- `test_helpers.rs`'s shared fixture seed now panics loudly via `.expect()` if boards were never loaded before it runs, instead of silently producing `active_board_id: None`.

`git grep -n 'loaded_or_empty' -- crates/kanban-tui/src` still returns hits, but only inside test fixtures in files outside this card's scope (`board_handlers.rs`, `column_handlers.rs`, `navigation_handlers.rs`, `sprint_handlers.rs`, `query.rs`), none of which were in the card's `files_to_touch` list. All production sites in the six files this card owns are migrated; see the notes below for the drift from the card's literal repo-wide grep claim.

## Notes for review

- The card's step 2 undercounted the per-frame skip tests as singular; there are two per-frame sites (`animation_tick.rs`, `sprint_management.rs`), each got its own test, as flagged in the handoff's corrections.
- One planned red test (`CardListAction::Complete` decline in `detail_view_handlers.rs`) was dropped: the `'c'` key that would dispatch `CardListAction::Complete` is shadowed by an earlier `KeyCode::Char('c')` arm in the same `handle_sprint_detail_key` match that always routes to `toggle_completion_for_card_ids` (multi-select) or `handle_complete_sprint_key` (single-select) instead, so the `CardListAction::Complete` match arm is unreachable dead code pre-existing this change. The production fix at that site is still correct in isolation (byte-equivalent Loaded branch, explicit non-Loaded branch), just not black-box testable via any key path today. Worth a follow-up card to either wire a reachable key to it or delete the arm.
- Several migrated sites (the carry-over-sprint label fallback, the relationship-popup search filters, `cursor_archive_anchor`, `handle_animation_tick`'s archive branch, `check_ended_sprints`'s board lookup) are behavior-preserving refactors: `loaded_or_empty()` on a `NotLoaded` tier already returned an empty slice, which is functionally identical to the new explicit skip branch. Confirmed via mutation-check (reverting each production fix and re-running its test): the acting-handler/banner sites all failed as expected when reverted; these skip-shape sites did not, since old and new code produce the same observable result. This matches the card's own SKIP-vs-DECLINE classification.
- Mutation-check was run per production site (revert fix, confirm test fails or is behavior-neutral, restore) rather than committed as a mutant; not included in the diff.

### DIFF
--- a/.changeset/kan-1498-tui-cards-boards-tier-decline.md
+++ b/.changeset/kan-1498-tui-cards-boards-tier-decline.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+tui: the 19 remaining `loaded_or_empty()` reads on the cards/boards tiers in kanban-tui's handlers and app modules now handle `LoadState` explicitly. Acting handlers decline with a "not loaded yet" banner and leave state unchanged when their cards or boards tier is not loaded; per-frame and read-only search paths skip silently; the shared test fixture that seeds `active_board_id` now panics loudly instead of silently producing `None` if boards were never loaded.
--- a/crates/kanban-tui/src/app/animation_tick.rs
+++ b/crates/kanban-tui/src/app/animation_tick.rs
@@ -1,5 +1,5 @@
 use super::{animation, App};
-use kanban_domain::AnimationType;
+use kanban_domain::{AnimationType, LoadState};
 use std::time::Instant;
 
 impl App {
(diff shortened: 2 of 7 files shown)


fulsomenko:
Addressed the review findings from attempt 2:

- `CardListAction::MoveColumn` no longer re-implements the collapsing `loaded_or_empty` read via a nested `_ => &[]` match; the cards slice is now bound once by the outer let-else guard and reused for both the card lookup and `compute_card_column_move`.
- All ten new decline tests now assert the banner variant and exact message text (via a shared `assert_error_banner` helper in each module) instead of only presence, so they can no longer pass on an unrelated banner from an earlier decline in the same call path.
- The `test_helpers.rs` panic-guard test was replaced: it now drives `setup_reload_resort_fixture` directly and asserts the seeded `active_board_id`, rather than re-typing the guarded expression inside the test body.

Ran `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test -p kanban-tui --all-features` (337 lib tests + integration suites), all green. Mutation-checked the banner-text assertions by temporarily corrupting a production error string and confirming the corresponding test failed, then restored it.

PR #840, opened 2026-09-11
KAN-1583 feat(tui): read handler operations from the board-scoped tiers

fulsomenko:
## Summary

Sweeps the remaining TUI handler reads of the three retiring flat `Model`
tiers (`cards_state`/`columns_state`/`sprints_state`) plus the global
`archived_card_ids()` reads onto board-scoped reads:

- Sprints and columns route through `App::board_sprints_view`/`board_columns_view`.
- Board-wide live card sets route through `Controller::live_cards()`/`archived_cards()`.
- Id-list card lookups route through `Model::card_by_id_state`.
- Create-card position derives from `Model::column_cards_state`.
- Archived-ness in relationship eligible-set building routes through a new
  `App::board_archived_ids(board_id)` helper instead of the global flat
  `archived_card_ids()`, which also repairs a currently-dead exclusion filter
  (a live relationship target now correctly excludes archived candidates in a
  `reload_model`-driven session, matching KAN-972's intent).
- `check_ended_sprints` now scopes to the active board instead of scanning
  the whole workspace's flat sprint tier.

### Pre-sweep census
48 production flat-tier lines (`cards_state`/`columns_state`/`sprints_state`)
plus 6 `archived_card_ids()` lines, via:
```
git grep -nE '\b(cards_state|columns_state|sprints_state)\(\)' -- crates/kanban-tui/src crates/kanban-view/src | grep -v archived_cards_state
git grep -n 'archived_card_ids()' -- crates/kanban-tui/src
```

### Post-sweep census
Same commands now return exactly seven production sites, all intentionally
out of this card's scope:

- `App::board_columns_view`'s `NotLoaded` fallback arm (query.rs), owned by the c04 ViewScope card (KAN-1585) that removes the fallback arms
- `App::board_sprints_view`'s `NotLoaded` fallback arm (query.rs), same
- `create_card_target_column`'s fallback arm (card_handlers.rs), same
- `handle_manage_children_from_list`'s column-set fallback arm (card_handlers.rs), same
- `prepare_frame`'s two flat gates (view_management.rs), owned by KAN-1585
- `tasks_panel_title`'s `all_tiers_loaded` gate (ui/main_view.rs), owned by KAN-1585

`archived_card_ids()` no longer appears outside `#[cfg(test)]`.

## Testing

Eleven new/updated tests pin the scoped-read behaviour:

- `test_handler_acts_on_scoped_tiers_while_flat_tiers_failed`
- `test_move_card_uses_the_scoped_column_and_partition_reads`
- `test_create_card_position_derives_from_the_column_tier`
- `test_toggle_completion_resolves_cards_by_id`
- `test_manage_children_eligible_set_spans_archived_when_target_archived`
- `test_manage_children_eligible_set_excludes_archived_when_target_live`
- `test_sprint_dialogs_read_the_board_scoped_sprint_tier`
- `test_check_ended_sprints_scopes_to_the_active_board` (inline, sprint_management.rs)

Three pre-existing test files needed fixture updates because they built the
model with `with_load_states`/`Model::with_load_states` or a wholesale
`app.model = ...` replacement, which never populates the scoped tiers or
resyncs the `Controller`'s card partitions: `detail_view_handlers_collapsing_accessor_tests.rs`,
`manage_children_parents_archived_candidates_tests.rs`, and the inline
`invalidate_cards_tier` helpers in `card_handlers.rs`/`detail_view_handlers.rs`
(which invalidated the model but never resynced the controller, leaving its
cached partitions stale rather than `NotLoaded`).

- `test_toggle_selected_cards_completion_with_a_failed_flat_cards_tier_for_an_unresolvable_id_declines`
- `test_toggle_completion_for_card_ids_with_a_failed_flat_cards_tier_for_an_unresolvable_id_declines`
- `test_relationship_search_with_failed_cards_tier_for_an_unresolvable_id_banners`

Behavioural fixes in round 2: `toggle_selected_cards_completion`,
`toggle_completion_for_card_ids` and `filtered_relationship_card_ids` now
decline the whole batch when a selected id's per-id lookup resolves `Failed`
(previously only `NotLoaded`), and the three eligible-set sites consolidate
onto the new `App::board_candidate_cards` helper in query.rs.

## Notes

- `check_ended_sprints` narrows its scope: it now reports `None` whenever no
  board is active, where previously it scanned every board's sprints
  regardless of which one (if any) was open. The three existing inline tests
  were updated to set an active board.
- `handle_manage_parents`/`handle_manage_children` now read `Controller::live_cards()`,
  which is scoped to whichever board `Controller::set_scope_board` was last
  called with. In production this always matches the card's own board (a
  card can only be active while its board is), but two test fixtures that
  set `active_card_id` without ever setting `active_board_id` needed an
  explicit `active_board_id` to keep the controller's scope aligned with the
  card under test.
- The `filtered_relationship_card_ids` and card-completion-toggle handlers
  keep their "not loaded" decline banners (an existing test pins that
  contract), now driven by `Model::card_by_id_state` per candidate instead
  of the flat collection.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --all-features --workspace`

### DIFF
--- a/.changeset/kan-1583-handler-scoped-reads.md
+++ b/.changeset/kan-1583-handler-scoped-reads.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+TUI handler operations read board-scoped tiers instead of the flat card/column/sprint collections
--- a/crates/kanban-tui/src/app/animation_tick.rs
+++ b/crates/kanban-tui/src/app/animation_tick.rs
@@ -24,7 +24,7 @@ impl App {
             self.animation.animating.remove(&card_id);
             match animation_type {
                 AnimationType::Archiving => {
-                    if let LoadState::Loaded(cards) = self.model.cards_state() {
+                    if let LoadState::Loaded(cards) = self.controller.live_cards() {
                         if let Some(card_pos) = cards.iter().position(|c| c.id == card_id) {
                             let card = &cards[card_pos];
                             if !affected_columns.contains(&card.column_id) {
--- a/crates/kanban-tui/src/app/card_selection.rs
+++ b/crates/kanban-tui/src/app/card_selection.rs
@@ -95,22 +95,16 @@ impl App {
             .and_then(|id| self.model.board_by_id_state(id).loaded().copied());
 
         let (uncompleted_ids, completed_ids) = if let Some(board) = board_opt {
-            if !self.model.columns_state().is_loaded() || !self.model.sprints_state().is_loaded() {
+            let LoadState::Loaded(columns) = self.board_columns_view(board.id) else {
                 self.set_error("Columns or sprints are not loaded yet".to_string());
                 return;
-            }
-            let columns = self
-                .model
-                .columns_state()
-                .loaded()
-                .expect("checked loaded above");
-            let sprints = self
-                .model
-                .sprints_state()
-                .loaded()
-                .expect("checked loaded above");
+            };
+            let LoadState::Loaded(sprints) = self.board_sprints_view(board.id) else {
+                self.set_error("Columns or sprints are not loaded yet".to_string());
+                return;
+            };
             let sorted_sprint_ids =
-                kanban_domain::CardQueryBuilder::new(cards, columns, sprints, board)
+                kanban_domain::CardQueryBuilder::new(cards, &columns, &sprints, board)
                     .in_sprints(std::iter::once(sprint_id))
                     .execute();
             let mut unc = Vec::new();
--- a/crates/kanban-tui/src/app/clipboard.rs
+++ b/crates/kanban-tui/src/app/clipboard.rs
@@ -11,14 +11,14 @@ impl App {
         if let Some(active_id) = self.selection.active_card_id {
             if let Some(board) = self.active_board() {
                 if let Some(card) = self.model.card_by_id_state(active_id).loaded().copied() {
-                    let LoadState::Loaded(sprints) = self.model.sprints_state() else {
+                    let LoadState::Loaded(sprints) = self.board_sprints_view(board.id) else {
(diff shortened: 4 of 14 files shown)


fulsomenko:
Addressed the three rejection findings:

1. Failed per-id resolution silently dropped a card instead of declining. `toggle_selected_cards_completion` (card_handlers.rs), `toggle_completion_for_card_ids` (detail_view_handlers.rs) and `filtered_relationship_card_ids` (popup_handlers.rs) now gate on `state.is_not_loaded() || state.is_failed()` instead of `is_not_loaded()` alone, so a Failed flat-tier resolution declines the whole batch with a banner rather than silently skipping the unresolvable id and acting on the rest. Kept `Missing` out of the gate on purpose: declining on Missing would be a new regression (a stale/deleted id should still be silently skipped, as before). Added a red test per site (id A resolves via the scoped tier, id B is unresolvable so it falls through to the Failed flat tier); confirmed each fails on the unmodified gate and passes with the fix by temporarily reverting the predicate and re-running.

2. Ticket reference in a code comment. Removed "named by KAN-1583" from the module doc comment in handler_scoped_reads_tests.rs; the rest of the doc comment (which states the file's test contract, not history) is unchanged.

3. Duplicated live-plus-archived candidate build across three sites. Extracted `App::board_candidate_cards()` in query.rs, returning `LoadState<(&[Card], &[Card])>` (live partition gates, archived partition degrades to empty exactly as before). `handle_manage_children_from_list`, `handle_manage_parents` and `handle_manage_children` all call it now instead of repeating the five-line archived-partition bind plus chain.

Verification: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all-features --workspace` all green.

PR #841, opened 2026-09-11
KAN-1585 refactor(tui,service,domain): stop requesting the flat column/card/sprint tiers

fulsomenko:
## Summary

- `ViewScope::next_round` stops requesting the workspace-flat `column_list`/`card_list`/`sprint_list` tiers. Against a backend that declines those whole-workspace calls (HttpBackend), the old F1 block re-requested them every single frame, since `requestable` treats `Failed` as retryable. That per-frame error loop is now gone.
- `board_sprints` becomes the sole sprint source: it moves into the base `ViewScope` (true by default in Normal/CardDetail/BoardDetail/SprintDetail/ArchivedBoardsView and every dialog with a board in scope) and is dropped only in Settings, unless a sprint filter is active. Every per-mode `scope.board_sprints = true` that duplicated the new base default is removed.
- `prepare_frame`'s task-list refresh and `tasks_panel_title`'s loaded-gate move from the flat `columns_state()`/`sprints_state()` accessors to the board-scoped `board_columns_state()`/`board_sprints_state()`. This completes the deviation reviewed in #839: search and the tasks panel now read the same scoped tiers the rest of the render path already used.
- The four dead scoped-then-flat-fallback code paths (`board_columns_view`, `board_sprints_view`, `create_card_target_column`, `handle_manage_children_from_list`) are removed. A scoped `Loaded` (including empty) is now the sole authoritative source; nothing falls back to the flat tier.
- `card_in_collection` retires from the `FetchPlan`/`LoadedState`/`LoadedEntities` vocabulary, along with `Model::card_in_collection_status` in kanban-domain. Its only production caller was already gone (the archived-card body walk reads the composed `card()` accessor as of #839/#840); this PR clears the stub plumbing and one surviving domain test reader.

## Corrections to the source card

- The card's Target step 3 described removing an F2 flat-merge block in the archived-card body walk. That block does not exist on `develop`; #839/#840 already rebuilt the walk onto `archived_cards_of_board`/`card()`. No change was needed there.
- The card estimated ~8 test rewrites. The actual blast radius, once the flat tiers stop being populated by `reload_model()`/`prepare_frame()`, was roughly 100 assertions across 23 integration-test files in `kanban-tui`, plus the refetch op-name assertions in `mutation_refetch_tests.rs` and `sprint_dialog_refetch_tests.rs`. Every migrated assertion keeps its original population path (still driven through the real handlers, never a snapshot load) and only changes which accessor it reads.

## Known gap (flagged, not fixed here)

`resolve_relationship_cards` (the CardDetail parents/children panel) resolves a related card through `Model::card_by_id_state`, which composes the per-id tier, the column-scoped tier, and (previously) the flat tier. A `spawns` relative that lives on a *different* board than the one currently in scope was, before this PR, incidentally resolvable because the flat tier held every board's cards. With the flat tier gone, a cross-board relative is no longer resolvable unless something has independently fetched its body.

`crates/kanban-tui/tests/relationship_resolution_tests.rs` pins the two tests that exercise this (`test_resolve_relationship_cards_resolves_a_cross_board_card_whose_body_was_fetched`, `test_card_detail_children_box_shows_a_cross_board_child_whose_body_was_fetched`) by seeding the cross-board card's body directly (a `warm_card_body` helper, the same shape as the existing `warm_archived_card_markers`). This keeps the tests honest about the resolver's contract, but it is not a production fix: CardDetail's `ViewScope` arm does not currently issue a per-id fetch for the graph's parent/child ids, so a real cross-board relative will show as missing from the panel today. Recommend a follow-up card to extend the `CardDetail` scope to request the active card's graph neighbours by id once the graph is loaded.

## Test plan

- `cargo test --workspace --all-features`: green.
- `cargo clippy --all-targets --all-features -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- Six new tests pin the behavior this PR introduces (scoped-only `next_round`, the base-scope `board_sprints` default and its Settings guard, `prepare_frame`/`tasks_panel_title` reading the scoped tiers under a failed flat tier, and a three-backend proof that the flat tiers stay `NotLoaded` while the scoped tiers load). Each was confirmed failing on the unmodified tree before implementation.

### DIFF
--- a/.changeset/kan-1585-viewscope-flat-arm-drop.md
+++ b/.changeset/kan-1585-viewscope-flat-arm-drop.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+stop requesting the flat column/card/sprint tiers; the render path and handlers already read the board-scoped ones
--- a/.changeset/kan-1585-viewscope-flat-arm-drop.md
+++ b/.changeset/kan-1585-viewscope-flat-arm-drop.md
@@ -2,4 +2,4 @@
 bump: patch
 ---
 
-stop requesting the flat column/card/sprint tiers; the render path and handlers already read the board-scoped ones
+tui: ViewScope stops requesting the flat column/card/sprint tiers; the render path and handlers already read the board-scoped ones. Consequence: CardDetail's parents/children panel no longer resolves a spawns relative on a different board unless its body was already fetched, since nothing populates the flat fallback any more.
--- a/crates/kanban-domain/src/model/apply.rs
+++ b/crates/kanban-domain/src/model/apply.rs
@@ -559,10 +559,11 @@ mod tests {
             assert_eq!(m.card_index[&card.id], i);
         }
         assert!(m
-            .card_in_collection_status(a.id)
+            .cards_state()
             .loaded()
-            .copied()
-            .is_none());
+            .unwrap()
+            .iter()
+            .all(|c| c.id != a.id));
         assert_eq!(m.card_by_id_state(a.id).loaded().copied().unwrap(), &a);
         assert_eq!(m.card_by_id_state(c.id).loaded().copied().unwrap(), &c);
         assert_eq!(m.card_by_id_state(d.id).loaded().copied().unwrap(), &d);
--- a/crates/kanban-domain/src/model/cards.rs
+++ b/crates/kanban-domain/src/model/cards.rs
@@ -87,24 +87,6 @@ impl Model {
             .unwrap_or(LoadState::NotLoaded)
     }
 
-    /// The flat collection tier alone, with no composition against the
-    /// per-id tier. Answers `NotLoaded` for an id absent from a `Loaded`
-    /// flat list rather than `Missing`, because `list_all_cards` excludes
-    /// archived cards by design: the id may still be requestable through a
-    /// per-id fetch even though the flat list has been read.
-    pub fn card_in_collection_status(&self, id: Uuid) -> LoadState<&Card> {
-        match self.cards.as_ref() {
-            LoadState::Loaded(cards) => {
-                match self.card_index.get(&id).and_then(|&idx| cards.get(idx)) {
-                    Some(card) => LoadState::Loaded(card),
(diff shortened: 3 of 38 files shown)


fulsomenko:
Addressed the three rejection findings.

1. Renamed the two relationship tests that manually warm a cross-board card's body (`test_resolve_relationship_cards_resolves_a_cross_board_card_whose_body_was_fetched`, `test_card_detail_children_box_shows_a_cross_board_child_whose_body_was_fetched`) so the name states the precondition instead of implying the production CardDetail path resolves a cross-board relative on its own. No production path fetches that body: `ViewScope`'s `CardDetail` arm only sets `scope.card`/`scope.graph`, and closing the gap needs a per-id graph-neighbour fetch that the shared fetch vocabulary (`LoadedEntities`/`ViewScope`) has no accessor for today, so it belongs to c05, not this card. Documented the gap as a follow-up on KAN-1585.

2. Restored the flat-only case to the three scoped-tier pinning tests (`test_board_columns_view_answers_not_loaded_when_the_scoped_tier_is_not_loaded`, `test_board_sprints_view_answers_not_loaded_when_the_scoped_tier_is_not_loaded`, `test_visible_board_columns_reads_only_the_scoped_tier`). Each now seeds only the flat tier with data for the target board and asserts the accessor still returns NotLoaded, so a reintroduced fallback would fail them again. Mutation-checked: reintroducing the fallback branch made the first of these three fail, then reverted.

3. Re-derived the startup regression guard's bound from the actual op count once the flat `ViewScope` arms are gone: `cargo test` shows exactly 2 whole-store list reads left (`list_all_cards` + `list_all_sprints`, both from `migrate_sprint_logs`). Tightened the assertion to `== 2` and rewrote the message to name only those two, dropping the reference to the now-deleted flat arms.

Full `cargo test --all-features --workspace`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo fmt --all -- --check` all green after the fixes.

PR #851, opened 2026-09-12
KAN-1595 docs(server,backend-http,tui): document conditional requests, pin LF endings, rebuild a tautological test

fulsomenko:
## Summary

Docs/infra/test-honesty slice of the KAN-1538/1541/1581 review harvest.

- Add a repo-root `.gitattributes` pinning `* text=auto eol=lf`. `git add --renormalize .` on a clean tree produced an empty diff, so nothing in the tree currently needs normalizing.
- Rebuild `test_resolve_relationship_cards_resolves_same_set_as_full_collection_scan` (renamed `test_resolve_relationship_cards_returns_live_archived_and_cross_board_bodies_in_input_order`), which built its expectation with the same `filter_map` the production code uses under the hood, so it could never fail. The new version asserts an independently derived literal order and adds an explicit check that the archived parent's `spawns` edge is absent from `graph.parents(subject)`, proving the archived id is resolved from the per-id tier rather than the graph. Mutation-checked by swapping two ids in the expected vector and confirming the assertion fails, then reverted before committing.
- Document the server's conditional-request protocol in `crates/kanban-server/README.md`: `ETag` construction, the exact routes that emit it, `If-None-Match`/304 semantics (including the weak-tag comparison), `If-Match`/412 semantics (including the absent-header short-circuit and the `*`-against-nothing failure), and the CORS caveat that only the explicit-origin-list mode exposes the header.
  - The route list for `If-Match` in this PR corrects the source card, which conflated `/v1/columns/{column_id}/cards/{id}` (which only registers `PUT`) with the board-scoped card routes (which register `PATCH`/`DELETE`).
  - While adding this section, found and removed a now-contradicted stale line: `Column writes (create/update/delete) aren't implemented yet.` The same README already lists the column write routes a few lines above, and this PR's new section documents `If-Match` on those same routes, so the sentence could not stay.
  - Corrected `GET /v1/graph`'s documented empty-graph body: it is `{"spawns":{"edges":[]},"blocks":{"edges":[]},"relates":{"edges":[]}}`, not `{}`.
  - Documented that a client-side `DependencyGraph` deserialization replays every edge through the same DAG cycle/self-reference guard the server enforces, with the archived-edge carve-out, and cross-referenced this from `crates/kanban-backend-http/README.md`.
- Added a module doc to `crates/kanban-backend-http/tests/sse_subscription.rs` recording the reconnect backoff and background-shutdown-on-drop behavior (no test added, this is a documentation-only note).
- Tightened three doc comments (`warm_archived_card_markers`, `warm_card_body`, `board_candidate_cards`) to state their real contracts rather than the vaguer prose they carried before.

## Deviation from the assigned card

The card's item 1 (a domain fix making `strip_test_modules` tolerate a trailing CR) was dropped. Its premise does not hold: Rust's `str::lines()` already treats `\r\n` as a single line terminator and strips both characters, so the cited sentinel check (`line == "}"`) already matches correctly on CRLF input. A red test built exactly as specified in the card passed at head with zero production changes, confirming there is no bug to fix. No change was made to `crates/kanban-domain/src/capabilities.rs`.

Two known-stale statements were left as-is per the card's scope, since nothing in this PR contradicts them: `crates/kanban-server/README.md:5` ("Only boards and column reads are wired up so far") and the sentence noting the remaining card routes aren't documented in the endpoint table yet.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test --all-features --workspace`
- [x] `capability_guard` (all four crates) and `write_lock_guard` remain green
- [x] Mutation-checked the rebuilt relationship-resolution test

### DIFF
--- a/.changeset/kan-1595-conditional-requests-docs-and-lf.md
+++ b/.changeset/kan-1595-conditional-requests-docs-and-lf.md
@@ -1,0 +1,5 @@
+---
+bump: patch
+---
+
+server: document the ETag / If-None-Match / If-Match protocol and the real /v1/graph empty body; repo: pin LF line endings; tui: rebuild a tautological relationship-resolution test on an independent expectation
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/crates/kanban-backend-http/README.md
+++ b/crates/kanban-backend-http/README.md
@@ -18,6 +18,10 @@ of running them through its local command-execute-then-log path.
 Every other `DataStore`/`CommandStore` method (graph, sprints, prefixes,
 archive/restore, the command log) still declines under its own name — see
 `test_http_backend_stub_method_returns_unsupported_error` in `src/lib.rs`.
+`HttpDataStore::get_graph` is the one read that does more than deserialize:
+it parses `GET /v1/graph` straight into the domain `DependencyGraph`, whose
+`Deserialize` impl validates the DAG on parse (see
+[the server README's Graph section](../kanban-server/README.md#graph)).
 Returning `Some` from `remote_writes()` also arms a service-layer fence
 (`KanbanContext::execute_with_extra`): every mutation this crate does not
 implement now fails fast with an explicit "not supported over the HTTP
--- a/crates/kanban-backend-http/README.md
+++ b/crates/kanban-backend-http/README.md
@@ -15,9 +15,12 @@ way, and `KanbanBackend::remote_writes()` returns `Some(self)`, so
 `KanbanContext` diverts those nine operations straight to the server instead
 of running them through its local command-execute-then-log path.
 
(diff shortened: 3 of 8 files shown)


fulsomenko:
Addressed both review findings:

1. Added the missing `GET /v1/columns/{id}` to the ETag-emitting GET enumeration in the Conditional requests section (all seven sites now listed, matching the already-correct If-Match list).
2. Corrected the SSE module doc's reconnect trigger from "reconnects on drop" to "reconnects whenever the event stream ends or the connection fails", removing the contradiction with the same sentence's `impl Drop` shutdown clause.

Checked for other instances of the same defect classes: no other route enumeration in the README omits a site, and no other doc uses "on drop" for the reconnect trigger.

PR #863, opened 2026-09-15
feat(backend-http): serve every whole-store list read by fanning out

fulsomenko:
## Summary

Closes the last gap in the HTTP-backend-parity series (#858-#862). Four `DataStore` reads declined as `Unsupported`, so **any unscoped listing failed against a remote `kanban-server`** — the MCP and CLI could only ever serve board-scoped queries:

- `list_all_cards`
- `list_all_columns`
- `list_all_sprints`
- `list_archived_cards`

Cards have no whole-store fetch tier in the domain model (`FetchRound` only has `board_list`/`archived_board_list`; cards are `cards_by_column` or by id). Rather than grow one — which would mean a new `Resolved` field, `FetchRound` field, `RouteScope` variant and resolve.rs wiring — each read is assembled client-side from per-board routes that already work. **No server or domain changes.**

## Two things worth review attention

**`all_board_ids` unions `/v1/boards` with `/v1/archived-boards`.** Archiving a board only records a marker; its columns and cards stay in the flat collections. A whole-store read that only walks live boards silently drops them. Pinned by `test_list_all_cards_over_http_includes_archived_board_descendants`.

**`list_all_columns` is load-bearing, not a bonus.** `list_live_cards_impl` calls `list_all_cards()` only when zero boards are archived; otherwise it needs `list_all_columns()` to exclude archived-board descendants. Implementing just the card reads would have left unscoped `card list` working until the first board was archived, then failing — a data-dependent break. Pinned by `test_unscoped_live_card_listing_survives_an_archived_board`.

## Regression guard

`sprint_management.rs` matched the operation name as a literal string (`operation: "list_all_cards"`). `migrate_sprint_logs` reads cards *then* sprints, so implementing `list_all_cards` alone would have advanced the decline to `list_all_sprints`, missed the arm, and restored the scary startup error entry #858 removed. Now matches `is_unsupported()`, which is immune to further decline-chain churn. This is `is_unsupported()`'s first production use — it existed but appeared only in tests.

## Semantics

Matched against the memory backend rather than invented:
- `list_all_cards` — live cards only (archived *cards* excluded), across live **and** archived boards, sorted by position/created_at/id
- `list_archived_cards` — all markers, ascending by `archived_at`
- `list_all_columns` — sorted by position/created_at/id
- `list_all_sprints` — sorted by `sprint_number`

Decline census 30 → 26; what remains is writes and single-entity marker reads.

## Test plan
- [x] `cargo test --all-features --workspace` — 304 test groups, 0 failed
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] 9 new tests in `whole_store_reads.rs` written red-first, incl. empty-store, cross-board aggregation, ordering, and both traps above
- [x] `test_unscoped_card_list_over_http_declines_under_list_all_cards` converted to assert success — it pinned a decline that no longer exists

KAN-1602

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CdmsHC6Ws9baThPqWaCrof


fulsomenko:
Rebased onto develop (047c47de). One conflict, in `tests/decline_rulings.rs`, resolved as follows.

Develop has moved under this branch since 2026-09-15:

- KAN-1606 implemented `count_cards_in_column_filtered`, so this branch's `test_count_cards_in_column_filtered_live_only_arm_declines_under_its_own_name` asserted a decline that no longer happens. Dropped, not merged.
- KAN-1604 made `get_archived_card` reach the transport, so it is already out of the census on develop. The branch's attempt to add it back was dropped.
- The `list_archived_cards` census entry is removed here, since this branch implements it.

Census count recomputed rather than guessed: develop was at 29 unconditional decliners, this branch implements four whole-store reads, so the assertion is now 25. Verified by running the suite, not by arithmetic: all 9 tests in `decline_rulings` pass.

The regression guard this card warned about is intact: `sprint_management.rs` now matches on `e.is_unsupported()` instead of the literal `operation: "list_all_cards"`, so advancing to the next declining read cannot resurrect the red-log bug that #858 fixed.

Local gates green: fmt clean, clippy `-D warnings` clean, kanban-backend-http and kanban-tui suites pass.

